PR #7, opened 2022-04-04
Make sure plugins are initialized

Philip-Scott:
This was one of the changes that I had to do in order to get the plugins working for me. Without this change, both the ExtraSettings and my music plugin were failing with errors similar to this one on the web console: 

```
Music:127 Uncaught (in promise) Plugin.get_player() missing 1 required positional argument: 'self'
```

### DIFF
--- a/plugin_loader/loader.py
+++ b/plugin_loader/loader.py
@@ -29,7 +29,7 @@ class Loader:
                 spec = spec_from_file_location("_", path.join(self.plugin_path, file))
                 module = module_from_spec(spec)
                 spec.loader.exec_module(module)
-                dc[module.Plugin.name] = module.Plugin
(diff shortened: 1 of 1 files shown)


marios8543:
The Plugin class is more of a container for all the methods and fields implemented by the plugin, and is not really meant to be instantiated. You got that error because you probably gave self as a positional argument in your methods, even though there isn't exactly a "self".
Although on hindsight it might be a good idea to instantiate them, for state-keeping and data sharing between methods.

marios8543:
Not merging, because this is already in the works along with other improvements to plugin loading. Thanks for bringing it to our attention though.

Philip-Scott:
Anytime! Let me know if you guys need help with anything :)

PR #13, opened 2022-04-04
Note it needing touchscreen input

LiamDawe:
As mentioned in issues, feels like an area that might need other things adding to it as issues are found :)


marios8543:
Yep. It's a good idea to mention it in the readme

PR #14, opened 2022-04-04
Update for new install instructions

LiamDawe:
Some changes to formatting to make it clearer, and I believe this is correct for the rewrite? Based on my testing.


marios8543:
Not merging because we're updating the README and install script to describe the installation of the auto-built binary, instead of manual installation. It's much more straightforward and easy to use

Issue #20, opened 2022-04-05
Plugins should be distributed as folders instead of single files


marios8543:
Use a folder structure such as
```
plugins/
     my_plugin/
           main.py
           tile_view.html
           main_view.html
```

PR #24, opened 2022-04-06
Feature/20 plugins as folders

ttay24:
First pass at implementing #20 

* changed the members in plugin to take in file names (see example below)
* updated file watcher to reload in `main.py` whenever a file underneath a plugin is created/modified
* note: if `tile_view_html` is left as an empty string, it will display the default (like it does today).

**Example updated plugin**
```python
class Plugin:
  # The name of the plugin. This string will be displayed in the Plugin menu
  name = "TTAY24 TEST"
  # The name of the plugin author
  author = "ttay24"

  # If the plugin should be reloaded from a call to /plugins/reload or a file change
  hot_reload = True

  # The HTML that will be loaded when selecting the plugin in the list
  main_view_html = "index.html"

  # The HTML that will be used to display a widget in the plugin main page
  tile_view_html = ""

  # A normal method. It can be called from JavaScript using call_plugin_function("method_1", argument1, argument2)
  async def method_1(self, *args):
    print("test")
    pass

  # A normal method. It can be called from JavaScript using call_plugin_function("method_2", argument1, argument2)
  async def method_2(self, *args):
    pass

  # Asyncio-compatible long-running code, executed in a task when the plugin is loaded
  async def __main(self):
    print("initializing plugin")
    pass
```
**Example directory structure**
image

We could probably expand on this to include pulling in a JS file, or we could update the `add_routes` to support pulling in files for plugins.

Open to any suggestions. I haven't really done much python dev, so any feedback is welcomed :) lol

### DIFF
--- a/plugin_loader/loader.py
+++ b/plugin_loader/loader.py
@@ -10,21 +10,45 @@ from logging import getLogger
 from injector import get_tabs
 
 class FileChangeHandler(FileSystemEventHandler):
-    def __init__(self, loader) -> None:
+    def __init__(self, loader, plugin_path) -> None:
         super().__init__()
+        self.logger = getLogger("file-watcher")
         self.loader : Loader = loader
+        self.plugin_path = plugin_path
 
     def on_created(self, event):
         src_path = event.src_path
         if "__pycache__" in src_path:
             return
-        self.loader.import_plugin(src_path, refresh=True)
+
+        # check to make sure this isn't a directory
+        if path.isdir(src_path):
+            return
+
+        # get the directory name of the plugin so that we can find its "main.py" and reload it; the
+        # file that changed is not necessarily the one that needs to be reloaded
+        self.logger.debug(f"file created: {src_path}")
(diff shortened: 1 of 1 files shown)


plugin_loader/loader.py
```diff
@@ -10,33 +10,58 @@
...
+        self.logger.debug(f"file modified: {src_path}")
+        plugin_dir = path.split(path.relpath(src_path, path.commonprefix([self.plugin_path, src_path])))[0]
+        self.loader.import_plugin(path.join(self.plugin_path, plugin_dir, "main.py"), plugin_dir, refresh=True)
 
 class Loader:
     def __init__(self, server_instance, plugin_path, loop, live_reload=False) -> None:
         self.loop = loop
         self.logger = getLogger("Loader")
         self.plugin_path = plugin_path
+        self.logger.info(f"plugin_path: {self.plugin_path}")
```
ttay24, line 58: Can remove this

marios8543:
Just tested this and it works great! Thanks for contributing :)

PR #32, opened 2022-04-07
Move all styles in styles.css

SpyrexDE:
Also adds some common classes for easier plugin UI development.


SpyrexDE:
I will close this because it will become rather unnecessary when components get implemented.

Issue #46, opened 2022-04-28
Uninstall instructions


ggppjj:
This plugin loader is really cool, and will likely get reinstalled at some point in the near future, but I'm looking to uninstall it while I wait out some of the more frustrating lockups that I've had with non-steam emulated games.

How would I uninstall the Plugin Loader from my Steam Deck, preferably without a reset?

Thanks!

Outpox:
This was asked a few days ago on the Discord server, here's the answer given by @marios8543:
```
sudo rm -rf ~/homebrew
sudo systemctl disable --now plugin_loader.service
sudo rm /etc/systemd/system/plugin_loader.service
```

ggppjj:
Thank you! Based on that, I've opened a pull request to add an uninstall script, #48. So this info is more readily available, I'd like to keep this issue open unless the pull request is accepted.

PR #48, opened 2022-04-28
Uninstall script addition

ggppjj:
Added uninstall script based on the answer provided to #46.

### DIFF
--- a/dist/uninstall.sh
+++ b/dist/uninstall.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-echo "Uninstalling Steam Deck Plugin Loader and all plugins..."
+echo "Uninstalling Steam Deck Plugin Loader..."
 
 HOMEBREW_FOLDER=/home/deck/homebrew
 
--- a/dist/uninstall.sh
+++ b/dist/uninstall.sh
@@ -1,3 +1,17 @@
-sudo rm -rf ~/homebrew
-sudo systemctl disable --now plugin_loader.service
-sudo rm /etc/systemd/system/plugin_loader.service
+#!/bin/sh
+
+echo "Uninstalling Steam Deck Plugin Loader and all plugins..."
+
+HOMEBREW_FOLDER=/home/deck/homebrew
+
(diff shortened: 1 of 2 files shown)


dist/uninstall.sh
```diff
@@ -0,0 +1,17 @@
...
+# Disable and remove services
+sudo systemctl disable --now plugin_loader.service > /dev/null
+sudo rm -f /home/deck/.config/systemd/user/plugin_loader.service
+sudo rm -f /etc/systemd/system/plugin_loader.service
+
+# Remove temporary folder if it exists from the install process
+rm -rf /tmp/plugin_loader
+
+# Remove folder structure
+sudo rm -rf ${HOMEBREW_FOLDER}
```
TrainDoctor, line 16:
```suggestion
# Cleanup services folder
sudo rm ${HOMEBREW_FOLDER}/services/PluginLoader
```
Before merging I think it would be best if the homebrew folder is preserved as if users are simply uninstalling to replace with a different version then it would be best to maintain the plugins folder etc.
I couldn't really get a nice safe ``rm`` invocation to clear the services folder nicely so this will need to be updated in the future as @WerWolv mentioned that there's the possibility of additional services so best to just purge specified contents of the folder for now.

TrainDoctor (changes requested):
Good changes, nice and simple but it's best to preserve the folder structure. Suggested change is needed as services folder doesn't need to be deleted but the contents should be uninstalled for users who want to re-install.

ggppjj:
Changes committed, although a full uninstall option would be helpful to consider at some point.

TrainDoctor:
> Changes committed, although a full uninstall option would be helpful to consider at some point.

While I agree, I think at this early in development making sure re-installs will work with no questions asked is a higher priority for the time being. And if users don't like the folder sticking around then they can delete them in desktop mode as needed.

Your point is very much taken and a goal should be set to have a full clear script either as an interactive option or just an alternative script. Thanks for your contribution!

@WerWolv and @marios8543 this looks good to merge.

PR #55, opened 2022-05-10
Use unique ids in call_server_method

### LINKED ISSUE #53: Race condition in call_plugin_method
Sometimes, when calling two Python functions at the same time from Javascript, the result of one will be returned by the other.
This is likely caused by the id of the calls being the same, since both are generated at (nearly) the same instant.

Rough reproduction/example:
```javascript
call_plugin_method("function1", {}).then(x => {
  document.getElementById("foo1").innerText = x.toString()
});
call_plugin_method("function2", {}).then(y => {
  document.getElementById("foo2").innerText = y.toString()
});```

### PR DESCRIPTION
**Pull request by patkub**
Tested this in Chrome console, not on Steam Deck yet.
```
// test
console.log(uuidv4());console.log(uuidv4());console.log(uuidv4());
// 08ad8294-06af-4f3c-9bfb-b1cace537df3
// ae64aa70-8d9d-4ef9-815f-c4198584eb1a
// 266f2f27-d774-4913-b984-8c4ddba004e6

console.log(new Date().getTime());console.log(new Date().getTime());console.log(new Date().getTime());
// 1652143359237
// 1652143359237
// 1652143359237
```

fixes #53

### DIFF
--- a/plugin_loader/static/library.js
+++ b/plugin_loader/static/library.js
@@ -8,7 +8,7 @@ window.addEventListener("message", function(evt) {
 }, false);
 
 async function call_server_method(method_name, arg_object={}) {
-    let id = `${new Date().getTime()}`;
+    let id = `${uuidv4()}`;
     console.debug(JSON.stringify({
         "id": id,
         "method": method_name,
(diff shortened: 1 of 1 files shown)


NGnius:
Looks like this would fix the race condition when that function is called too quickly as well. Thanks for putting in the work!

PR #58, opened 2022-05-13
Switch frontend rendering to react via hook

JohnnyCrazy:
Successor of #41


wvffle (commented):
I am not a react developer though I found some things that could be optimized.

PR #61, opened 2022-05-18
Increase User Security by firewalling the remote CEF port by default

Wolvan:
By default, once CEF is enabled, in addition to the `localhost:8080` CEF is also available on `0.0.0.0:8081`. As this is potentially insecure, as everyone on the network would be able to interact with a user's deck, PluginLoader now disables access to the port 8081 from anything but `localhost` via iptables by default.
For developers, the port can be re-enabled by placing an empty file with the name `allow_cef_from_network` in the same directory as the `PLUGIN_PATH` folder (usually `~/homebrew`).
PluginLoader will pick up on this and skip firewalling off the CEF port.

### DIFF
--- a/plugin_loader/main.py
+++ b/plugin_loader/main.py
@@ -40,6 +40,12 @@ async def chown_plugin_dir(_):
     Popen(["chown", "-R", "deck:deck", CONFIG["plugin_path"]])
     Popen(["chmod", "-R", "555", CONFIG["plugin_path"]])
 
+async def disable_cef_remote_port(_):
+    cef_firewall_disable_file = path.join(CONFIG["plugin_path"], "../allow_cef_from_network")
+    if (not path.exists(cef_firewall_disable_file) or not path.isfile(cef_firewall_disable_file)):
+        logger.info("Firewalling CEF to deny remote connections")
+        Popen(["iptables", "-I", "INPUT", "!", "-i", "lo", "-p", "tcp", "--dport", "8081", "-j", "DROP"])
+
 class PluginManager:
     def __init__(self) -> None:
         self.loop = get_event_loop()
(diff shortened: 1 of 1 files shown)


WerWolv:
There's a service running on the Deck called `steam-web-debug-portforward` which is responsible for forwarding the local CEF debugging port `:8080` to the network on port `:8081`. We can simply stop that service to prevent the port from being forwarded.

That might be a bit easier to implement and deal with ultimately than having to resort to iptables. What's your opinion on that?

Wolvan:
> > There's a service running on the Deck called `steam-web-debug-portforward` which is responsible for forwarding the local CEF debugging port `:8080` to the network on port `:8081`. We can simply stop that service to prevent the port from being forwarded.
> > That might be a bit easier to implement and deal with ultimately than having to resort to iptables. What's your opinion on that?
> 
> My only follow up would be that we need to determine how that process is launched so we can start it again ourselves. Sorry if the solution is super obvious but I'm not in a capacity to investigate at the moment.

It's a normal systemd service, so we just use systemctl to start and stop it

botatooo:
Also, consider replacing `subprocess.Popen` by `subprocess.call`

TrainDoctor:
> Also, consider replacing `subprocess.Popen` by `subprocess.call`

I'm going to make this suggestion a requirement due to the previously discovered zombie processes.

Titaniumtown:
This is probably a good idea to merge! Don't want random people to connect into my steam deck lol

pastaq:
This also doesn't impede remotely accessing CEF via authentication if you set up you dev environment correctly. By using an ssh tunnel you can map port 8080 to a remote machine port and access it via devtools as localhost:PORT on that workstation.

Example(run from remote workstation): 
`ssh $(SSH_USER)@$(SSH_HOST) -N -f -L 4040:localhost:8080`

TrainDoctor:
Supplanted by #129.

PR #62, opened 2022-05-19
Sort Plugins by directory modification time

milroneth:
User can order plugins by installing/updating them in the order they
want them to, instead of them being displayed in arbitrary order.


TrainDoctor:
This will be closed and converted to an issue to be implemented in the new settings page.

PR #68, opened 2022-05-30
react: Add Router hook & fix typescript issues

AAGaming00:
Adds an API for adding our own routes to the `react-router` instance.

Also fixes issues with TypeScript's JSX support.


AAGaming00:
CI fails because decky-frontend-lib hasn't been updated on NPM.

PR #70, opened 2022-05-30
Update README.md, password is needed

LiamDawe:
There is no password by default, so people need to set one before running that script.


TrainDoctor (commented):
Personally I'd include a link to your video on how to set your password in desktop mod. I would approve of this shameless plug. 😝

LiamDawe:
Oh okay :), I've added it.

TrainDoctor:
Got approval to merge this. Thanks for your contribution @LiamDawe!

PR #111, opened 2022-07-06
Refractor plugin backend

botatooo:
Use server methods instead of directly hitting up the browser endpoint, makes it less janky.


AAGaming00 (changes requested):
Maybe do this for the install method too?

botatooo:
Plugin installation already utilizes server methods, no?

botatooo:
Oh, I see what you mean.

AAGaming00:
@WerWolv opinions on the formatting applied here?

AAGaming00 (approved):
Code-wise LGTM, still waiting on Wolv's opinion on the formatter. Will test soon.

WerWolv:
Code-wise everything looks alright to me as well.

Formatting-wise, it partially feels like you just used an auto formatter that only allows a certain number of characters per line.
For me this makes it somewhat harder to read. Array indexes in my opinion should never be split up onto more than one lines. If the index value is too long, I'd write the result to a variable and then use that to index.
For function calls, if there are way too many parameters, I'll usually put each parameter on its own line and not just put the entire list of parameters on the next line.

botatooo:
I'll take a note and do what I can in the next commit.

botatooo:
> Code-wise everything looks alright to me as well.
> 
> Formatting-wise, it partially feels like you just used an auto formatter that only allows a certain number of characters per line. For me this makes it somewhat harder to read. Array indexes in my opinion should never be split up onto more than one lines. If the index value is too long, I'd write the result to a variable and then use that to index. For function calls, if there are way too many parameters, I'll usually put each parameter on its own line and not just put the entire list of parameters on the next line.

Did what I could manually, if it needs changes i will revisit it with a proper editor and not GitHub mobile.

TrainDoctor (approved):
This is good for merging.

Gerrickjohnson (approved):

AAGaming00:
This needs to be adapted for recent changes.

botatooo:
> This needs to be adapted for recent changes.

Conflicts have been resolved.

TrainDoctor (changes requested):
Tested it on my deck via manual file replacement and works like a charm.
Did encounter an issue where the uninstall modal shows up but neither confirm or cancel are selected. Hitting the "A" button to confirm as if confirm was selected seems to press the uninstall button behind the modal and uninstall the plugin as if confirm was selected.
Once this issue is resolved, this will be good to merge.

TrainDoctor (approved):
With recent discoveries testing @pastaq's changes the uninstall prompt issues appear to be in the main codebase so we'll look into sorting that ourselves. I'll need another approving review before this can fully go ahead.

AAGaming00 (approved):
LGTM, will test when I have time.

AAGaming00 (approved):

AAGaming00 (approved):

backend/main.py
```diff
@@ -80,9 +94,7 @@ def __init__(self) -> None:
         self.web_app.add_routes([get("/auth/token", self.get_auth_token)])
 
         for route in list(self.web_app.router.routes()):
-             self.cors.add(route)
```
TrainDoctor: Why where these lines removed?

backend/utilities.py
```diff
@@ -81,19 +81,15 @@ async def execute_in_tab(self, tab, run_async, code):
         try:
             result = await inject_to_tab(tab, code, run_async)
             if "exceptionDetails" in result["result"]:
-                return {
```
TrainDoctor, line 70: Why where these lines removed?

TrainDoctor (changes requested):
A few line removals that need to have their removal checked

TrainDoctor:
> A few line removals that need to have their removal checked

Also a few errors relating to indentation and so on causing the build CI to fail.

AAGaming00 (approved):
Tested locally, everything seems to work.

PR #123, opened 2022-07-29
Use Environment Variables

pastaq:
- Uses environment variables instead of hard coding the "deck" user/group. This
adds support for systems other than the steam deck that are using the DeckUI.

### DIFF
--- a/backend/browser.py
+++ b/backend/browser.py
@@ -43,8 +43,8 @@ class PluginBrowser:
             return False
         zip_file = ZipFile(zip)
         zip_file.extractall(self.plugin_path)
-        code_chown = call(["chown", "-R", get_user()+":"+get_user_group(), self.plugin_path)
-        code_chmod = call(["chmod", "-R", "555", self.plugin_path)
+        code_chown = call(["chown", "-R", get_user()+":"+get_user_group(), self.plugin_path])
+        code_chmod = call(["chmod", "-R", "555", self.plugin_path])
         if code_chown != 0 or code_chmod != 0:
             logger.error(f"chown/chmod exited with a non-zero exit code (chown: {code_chown}, chmod: {code_chmod})")
             return False
--- a/backend/browser.py
+++ b/backend/browser.py
@@ -7,9 +7,9 @@ from io import BytesIO
 from zipfile import ZipFile
 from concurrent.futures import ProcessPoolExecutor
 from asyncio import get_event_loop
-from time import time
+from time import time, sleep
 from hashlib import sha256
-from subprocess import Popen, check_output
+from subprocess import Popen, check_output, PIPE
 from injector import inject_to_tab
 
 import json
@@ -41,16 +41,22 @@ class PluginBrowser:
             return False
         zip_file = ZipFile(zip)
         zip_file.extractall(self.plugin_path)
+
         USER = None
         cmd = "who | awk '{print $1}' | sort | head -1"
+        # Get the user by checking for the first logged in user. As this is run
+        # by systemd at startup the process is likely to start before the user
+        # logs in, so we will wait here until they are available. Note that
+        # other methods such as getenv wont work as there was no $SUDO_USER to
+        # start the systemd service.
         while USER == None:
-            USER_LIST = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout = subprocess.PIPE, stderr = subprocess.PIPE, shell=True)
+            USER_LIST = Popen(cmd, stdin=PIPE, stdout=PIPE, stderr=PIPE, shell=True)
             for get_first in USER_LIST.stdout:
                 name = get_first.decode().strip()
                 if name is not None:
                     USER = name
                     break
-            time.sleep(1)
+            sleep(1)
         GROUP = check_output(["id", "-g", "-n", USER]).decode().strip()
         Popen(["chown", "-R", USER+":"+GROUP, self.plugin_path])
         Popen(["chmod", "-R", "555", self.plugin_path])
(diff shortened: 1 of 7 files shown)


backend/browser.py
```diff
@@ -41,7 +41,18 @@ def _unzip_to_plugin_dir(self, zip, name, hash):
...
+        USER = None
+        cmd = "who | awk '{print $1}' | sort | head -1"
+        while USER == None:
+            USER_LIST = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout = subprocess.PIPE, stderr = subprocess.PIPE, shell=True)
+            for get_first in USER_LIST.stdout:
+                name = get_first.decode().strip()
+                if name is not None:
+                    USER = name
+                    break
+            time.sleep(1)
```
TrainDoctor, line 53: As already discussed this call to time.sleep(1) seems to break based upon how imported the time module. More testing can be done once this is resolved.

TrainDoctor (changes requested):
Overall good changes that help to support an array of hardware rather than just a hardcoded deck user but needs the mentioned bug fix along with testing on both real Steam Deck hardware and on desktop linux as needed. Thanks for your PR and hopefully we can help you to get this running.

backend/browser.py
pastaq: Changed this and identified a few other import issues as well, specifically i was using the subprocess prefix and I missed the PIP import. Also used the asyncio sleep in main.py instead of time wince that is an async function which I missed the first push. Everything should be testable now.

TrainDoctor (approved):
Looks good to go. Turns out the uninstall prompt issues are in our code base so sorry @botatooo I'll be reviewing your code in that context to asap. I'll need another approving review from others before this can fully go ahead.

AAGaming00 (changes requested):
The code for determining the current user should be made into a utility instead of duplicated across files.

backend/browser.py
```diff
@@ -41,7 +41,24 @@ def _unzip_to_plugin_dir(self, zip, name, hash):
             return False
         zip_file = ZipFile(zip)
         zip_file.extractall(self.plugin_path)
-        Popen(["chown", "-R", "deck:deck", self.plugin_path])
+
+        USER = None
+        cmd = "who | awk '{print $1}' | sort | head -1"
+        # Get the user by checking for the first logged in user. As this is run
```
TrainDoctor, line 47: Duplication of code. Please move to a utility method that can be invoked in browser and main.

backend/main.py
```diff
@@ -32,7 +33,23 @@
 logger = getLogger("Main")
 
 async def chown_plugin_dir(_):
-    code_chown = call(["chown", "-R", "deck:deck", CONFIG["plugin_path"]])
+    USER = None
+    cmd = "who | awk '{print $1}' | sort | head -1"
+    # Get the user by checking for the first logged in user. As this is run
```
TrainDoctor, line 38: As previously stated duplicated code.

backend/main.py
```diff
@@ -32,7 +33,23 @@
 logger = getLogger("Main")
 
 async def chown_plugin_dir(_):
-    code_chown = call(["chown", "-R", "deck:deck", CONFIG["plugin_path"]])
+    USER = None
+    cmd = "who | awk '{print $1}' | sort | head -1"
```
TrainDoctor, line 37: Relating to previous issues found with .Popen, this terminal invocation would preferably be a set of python code. If not possible to use python then measures to ensure the prevention of a zombie process from .Popen should be taken.

TrainDoctor (changes requested):
Upon further review and previous issues encountered with earlier versions relating to .Popen causing zombie processes we'd like to request that the terminal invocations be moved over to python code where possible. We are flexible if this isn't viable but more code to ensure the prevention of zombie processes will be required. And of course, as @AAGaming00  already stated it would be helpful if the code for finding the user be moved into a utility function in helpers.py.

Also, if possible caching the result of this would be desirable so that the function doesn't need to be ran after first boot. (This is desired but not required.)

backend/browser.py
pastaq: Done

backend/main.py
pastaq: Done

backend/main.py
pastaq: All code now uses call or check_output.

pastaq:
Made requested changes. Also noticed some inefficiencies/duplication in imports so I ordered them alphanumerically and simplified when they were duplicated, moved all logic to after imports.

backend/main.py
```diff
@@ -1,10 +1,29 @@
...
+from browser import PluginBrowser
+from helpers import get_user, get_user_group
+from injector import inject_to_tab, tab_has_global_var
+from loader import Loader
+from updater import Updater
+from utilities import Utilities
 
+USER = get_user()
+GROUP = get_user_group()
+HOME_PATH = "/home/"+USER
```
JakobDev, line 24:
```suggestion
HOME_PATH = os.path.expanduser("~")
```
It is possible to have a home directory outside /home

TrainDoctor (approved):
This appears good to merge. I'm waiting for testers to provide additional feedback and once they have provided me with feedback, I can proceed with merging if no problems are found during testing.
Thank you again for your time working on this.

backend/main.py
TrainDoctor: While this is technically correct, we are focused on supporting primarily Steam Deck hardware + SteamOS 3. We will take steps to help support other distros but if their home directories do not exist withing ``/home`` then that's going to be exotic enough that we're not going to spend time supporting that. Apologies.

TrainDoctor:
Currently this doesn't account for multiple users but since the deck is a single user (at a time) system this is a non-issue. If issues do arise on non-deck systems relating to multiple users being logged in, PRs will be accepted to resolve the issue.
Moving to merge this into main. Thanks for your contribution @pastaq. Sorry for giving you the run around for so long 😛

PR #138, opened 2022-08-12
Remove console.log that was causing lots of log spam

OMGDuke:
I don't know if this console.log was added for a reason but it get called a lot of fills the log with various React Nodes.

Please close if the log is on purpose!


AAGaming00:
Oops, missed this entirely.
Thanks for the contribution!

AAGaming00 (approved):

Issue #102, opened 2022-07-03
Loader - Automatic Update Functionality


TrainDoctor:
Checklist for implementing update notifications and an updater into the loader:

CI:
- [x] CI successfully identifies a new full release and can create a new pre-release w/ a minor iteration via semver.
- [x] CI can create a new full release from last release to now (not-required for react update full-release)

Backend:
- [x] Method to check if the current installation of the loader is up to date, and return the current version, latest version, and patch notes to the frontend
- [x] Method to update the current installation of the loader, restart it, and reload SP (without restarting Steam)
- [x] Underlying regex to check which build version is being used from the modal (check for "nightl" for nightlies and "pre" etc)

Frontend:
- [x] Check on startup and periodically (every couple hours) with the backend for a new available version
- [x] Notify the user with a toast notification that there is a new version available
- [x] Modal to select between stable, pre-release and nightly builds
- [x] A button-text combo in general settings to:
     - [x] Show the latest and current version of the loader.
     - [x] Offer a button to trigger an update if one is available
     - [x] Show patch notes

PR #168, opened 2022-09-01
Addition of proper branch slection

TrainDoctor:
Addresses elements of #102.


TrainDoctor:
Thanks to @AAGaming00 for figuring what in the typescript mess was causing the branch select element not to update proper.

PR #170, opened 2022-09-05
Auto-update CI stuffs

TrainDoctor:
Major release revisions should still be done manually though, but in a pinch these should make life much easier.

This will address the final elements of #102 and that issue can be marked done once this is merged.


TrainDoctor:
Supplanted by recent commits. All major or minor releases should be done by hand, although going from a release from another previous release will iterate the minor version.

PR #176, opened 2022-09-15
Add note about opening terminal to README

saamirt:
Quick note about how to open the terminal using the pre-installed Konsole app for users that are either new to linux or Steam Deck.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Keep an eye on the [Wiki](https://deckbrew.xyz) for more information about Plugi
 4. Under Miscellaneous, enable `CEF Remote Debugging`
 5. Click on the `STEAM` button and select `Power` -> `Switch to Desktop`
 6. Make sure you have a password set with the "passwd" command in terminal to install it ([YouTube Guide](https://www.youtube.com/watch?v=1vOMYGj22rQ)).
-7. Open a terminal and paste the following command into it:
+7. Open a terminal ("Konsole" is the pre-installed terminal application) and paste the following command into it:
     - For the latest pre-release:
         - `curl -L https://github.com/SteamDeckHomebrew/decky-loader/raw/main/dist/install_prerelease.sh | sh`
     - For testers/plugin developers:


TrainDoctor:
While I appreciate the tweak, most users will not be using a separate terminal application and if they are, their steam deck is already in a super non-standard configuration.

vrman123:
I personally think this tweak would be helpful, as some users don't know what the terminal is at all, and so some mention of konsole would make it easier to understand

saamirt:
> While I appreciate the tweak, most users will not be using a separate terminal application and if they are, their steam deck is already in a super non-standard configuration.

I'm not sure I get what you mean. Konsole is the default preinstalled terminal application afaik. I'm not sure what else you would expect most users to use for a terminal.

enbyautumn:
Quite a few people might not know that Konsole is the terminal app, as people who have not used KDE before likely will not

TrainDoctor:
If you type terminal into the search konsole pops up because kde understands that there might be confusion. I know that users might not know the exact name of the program but I'm sure uaing their intelligence they can piece the two together.

saamirt:
Keep in mind that most users have not used a terminal, and likely do not know what that is. They would not search for "terminal", they would google the phrase and spend 5 minutes looking around to find instructions. Even if they did search it though, this would confirm that Konsole is what they're expected to use rather than being half-sure of the search result that comes up.

This small note can only ever help, not hurt. I honestly didn't expect this change to be contentious.

vrman123:
People often struggle on this specific step and so I have often had to explain how to open the terminal on the steam deck. This change could only be useful, I'm not totally clear why you would assume users can figure it out on their own. Every step of difficulty is a step towards them giving up, as most users just want it to *work* with as little thought as possible.

TrainDoctor:
> People often struggle on this specific step and so I have often had to explain how to open the terminal on the steam deck. This change could only be useful, I'm not totally clear why you would assume users can figure it out on their own. Every step of difficulty is a step towards them giving up, as most users just want it to _work_ with as little thought as possible.

A fair point. Apologies @saamirt being a linux user for long enough has distorted my perception of what is and is not obvious to new users.

saamirt:
I'm not sure I get what you mean. Konsole is the default preinstalled
terminal application afaik. I'm not sure what else you would expect most
users to use for a terminal.

PR #177, opened 2022-09-17
Download Remote Binaries during Store Install

joamjoamjoam:
Download Remote Binaries during store install.
Fix Manual Install Modal Type

Note: will need to remove CI changes for remote binary so they don't download and bundle remote binaries. Once this is approved I can submit those changes also.

Link for Installing RemotePlayWhatever for testing. You can use the Manual Install via web to test it.
https://github.com/joamjoamjoam/RemotePlayWhatever/releases/download/testRemoteBinary/RemotePlayWhatever.zip


joamjoamjoam:
Fixed Line Endings

TrainDoctor (approved):

PR #178, opened 2022-09-19
Grab Folder Name instead of .zip name

joamjoamjoam:
Grab the plugin path using find_plugin_folder since the plugin name is not the plugin folder name.


joamjoamjoam:
Tested CI built service with multiple zips and plugin names. Works as expected.

PR #185, opened 2022-09-25
fix(pluginview): align icon and plugin name

FrogTheFrog:


TrainDoctor:
This PR will be reviewed at ASAP, but will only be merged affter the publishing of the next stable release. Thanks for your patience.

PR #190, opened 2022-10-01
Change 'PluginLoader' to 'Decky Loader' in readme

PartyWumpus:
This seems like a reasonable change as it hasn't been called PluginLoader for a while.

Would it have been better to make an issue suggesting the change instead or is this fine?


TrainDoctor (approved):
🤦🏻

Issue #197, opened 2022-10-04
Plugin menu not loading


Wintersdark:
Installed as per directions.  Sudo password set and used, install appears to proceed without problems.  

Developer mode on, CEF remote debugging enabled.  I am not running Crankshaft, or Syncthing.

When I reboot, there's no plugin menu on the quick access menu. 

journalctl -u plugin_loader.service outputs:
```


Oct 03 20:20:58 steamdeck systemd[1]: Started SteamDeck Plugin Loader.
Oct 03 20:20:59 steamdeck PluginLoader[22822]: main.py:63: DeprecationWarning: There is no current event loop
Oct 03 20:20:59 steamdeck PluginLoader[22822]: [loader][INFO]: plugin_path: /home/deck/homebrew/plugins
Oct 03 20:20:59 steamdeck PluginLoader[22822]: [updater][INFO]: /home/deck/homebrew/services
Oct 03 20:20:59 steamdeck PluginLoader[22822]: [updater][INFO]: Current branch is not set, determining branch from version...
Oct 03 20:20:59 steamdeck PluginLoader[22822]: [updater][INFO]: Current version determined to be pre-release
Oct 03 20:20:59 steamdeck PluginLoader[22822]: [loader][INFO]: import plugins from /home/deck/homebrew/plugins
Oct 03 20:21:30 steamdeck PluginLoader[22822]: [updater][INFO]: Current branch is not set, determining branch from version...
Oct 03 20:21:30 steamdeck PluginLoader[22822]: [updater][INFO]: Current version determined to be pre-release
Oct 03 20:21:30 steamdeck PluginLoader[22822]: [updater][INFO]: Updated remote version information
Oct 03 20:21:42 steamdeck PluginLoader[22822]: [main][INFO]: Plugin loader isn't present in Steam anymore, reinjecting...
Oct 03 20:22:02 steamdeck PluginLoader[22822]: [main][INFO]: Plugin loader isn't present in Steam anymore, reinjecting...
Oct 03 20:22:02 steamdeck PluginLoader[22822]: [main][INFO]: Failed to inject JavaScript into tab
Oct 03 20:22:02 steamdeck PluginLoader[22822]: Traceback (most recent call last):
Oct 03 20:22:02 steamdeck PluginLoader[22822]:   File "main.py", line 127, in inject_javascript
Oct 03 20:22:02 steamdeck PluginLoader[22822]:   File "injector.py", line 91, in inject_to_tab
[...]

PR #201, opened 2022-10-04
Fixed an issue regarding the scroll root path

proEndreeper:
The scroll root's path changed in the latest Steam Client beta update. This PR makes so that Decky should be able to find the scroll root using a recursive function that goes down the sibling and child nodes. This fixes issue #197 from what I can tell. Tested personally to work on both the latest Beta update, and the stable update, and works on both.

### DIFF
--- a/frontend/src/tabs-hook.ts
+++ b/frontend/src/tabs-hook.ts
@@ -47,18 +47,24 @@ class TabsHook extends Logger {
     const self = this;
     const tree = (document.getElementById('root') as any)._reactRootContainer._internalRoot.current;
     let scrollRoot: any;
-    let currentNode = tree;
+    async function findScrollRoot(currentNode: any, iters: number): Promise<any> {
+      if (iters >= 30) {
+        return null;
+      }
+      currentNode = currentNode?.child;
+      if (currentNode?.type?.prototype?.RemoveSmartScrollContainer) return currentNode;
+      if (!currentNode) return null;
+      if (currentNode.sibling) {
+        let node = await findScrollRoot(currentNode.sibling, iters++);
+        if (node !== null) return node;
+      }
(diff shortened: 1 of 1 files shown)


TrainDoctor:
@SteamDeckHomebrew/decky-testers Please test this and provide feedback ASAP!

frontend/src/tabs-hook.ts
```diff
@@ -47,18 +47,24 @@ class TabsHook extends Logger {
     const self = this;
     const tree = (document.getElementById('root') as any)._reactRootContainer._internalRoot.current;
     let scrollRoot: any;
-    let currentNode = tree;
+    async function findScrollRoot(currentNode: any, iters: number): Promise<any> {
+      if (iters >= 30) {
+        return null;
```
AAGaming00, line 52: This is where we would sleep and retry instead of returning null.

frontend/src/tabs-hook.ts
```diff
@@ -72,7 +78,7 @@ class TabsHook extends Logger {
                 this.tabRenderer = ret.props.children[1].children.type;
                 newQATabRenderer = (...args: any) => {
                   const oFilter = Array.prototype.filter;
-                  Array.prototype.filter = function (...args: any[]) {
```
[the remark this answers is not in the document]
proEndreeper: This was from prettier (using `npm run format`)

frontend/src/tabs-hook.ts
proEndreeper: I have made the changes you requested in a new commit, please review when you have a moment.

AAGaming00 (approved):
Code looks good, awaiting review from testers.

TrainDoctor (approved):
Reviewing on the behalf of @PartyWumpus and discord user Ollum, verified working on preview.

PR #208, opened 2022-10-07
README: password typing clarification

PartyWumpus:
While the video does *show* the password not appearing, he doesn't mention it. This should reduce the occasional user confusion of passwords not appearing.


PartyWumpus:
image

Issue #203, opened 2022-10-04
v2.2.2-pre1 Borked Quick Access?


GeneStarwind09:
I upgraded to the latest prerelease and Decky vanished from quick access. I tried a reinstall and nothing. I'll try switching to stable tomorrow but just figured I'd give you a heads-up.

TrainDoctor:
v2.2.2-pre1 is indentical to v 2.2.1 on stable.

GeneStarwind09:
I tried an uninstall of Decky and then an install of stable. I can see that it installed because I get the notification upon boot saying an update is available. However, it's not in quick access.

GeneStarwind09:
I uninstalled Decky and then installed the stable version. I can tell that it installed because I get the notification saying there's an update. However, it isn't in quick access.

TrainDoctor:
Please restart your deck if you haven't already. Just in case.

PartyWumpus:
If you're still getting the notification that there's an update you're probably not on the newest version? Try reinstalling/updating by tapping on the notification

TrainDoctor:
According to testers, this issue should be fully resolved on pre-release v2.2.3-pre1. Please install this version via the pre-release install script if you can't get the update notification. @re11ding and @GeneStarwind09.

re11ding:
It worked! It's no longer disappearing from what I can tell. Thanks for fixing that up!

GeneStarwind09:
Worked for me too, cheers.

PR #212, opened 2022-10-08
Use a conditional while loop for when the scroll root is not found.

proEndreeper:
After diagnosing the issue regarding Decky not loading in select situations but printing the error that states it was unable to find scroll root, it appeared to be caused by the recursive function never hitting 30 levels of depth, thus it never rechecks the tree. This seems to be partially caused by TabsHook loading before the QAM has been initialized in Library Init Stage 2, which occurs randomly on occassion, and seems to be less likely to occur on Steam Decks that don't have the lock screen setup. The proposed fix is a revert of some portion of the code to where a while loop is used after the initial call attempting to get the scroll root instead of just exiting, as well as removing the portion that would continue the recursive function should 30 recursive calls be made. 

On another note, I have added some log messages regarding the recursion limits. One denotes the current level that it hits when it finds the scroll root, the other denotes that the recursion limit needs to be increased, as the max depth of the tree is now longer than the recursion call, but this should not appear currently since currently the scroll root is only around 23 levels deep right now, and the current limit is 30, which is still above the depth of the tree. The most recent update that broke Decky and is why the changes were made to how the scroll root was found were made changed the relative level from around 21 to around 23, so it may be possible that the scroll root will be on a level deeper than 30 someday.

### DIFF
--- a/frontend/src/tabs-hook.ts
+++ b/frontend/src/tabs-hook.ts
@@ -72,7 +72,6 @@ class TabsHook extends Logger {
         this.log('Failed to find scroll root node, reattempting in 5 seconds');
         await sleep(5000);
         scrollRoot = await findScrollRoot(tree, 0);
-        return;
       }
       let newQA: any;
       let newQATabRenderer: any;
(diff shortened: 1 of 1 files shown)


proEndreeper:
This should fix issue #203 , which was a symptom of the TabsHook failing to inject ocassionally, which appears to happen reliably if Decky loads prior to Library Init Stage 2.

AAGaming00 (approved):

nabel0 (approved):
Installed and rebootet several times...Each time it was working fine.

PR #239, opened 2022-10-29
Add timeout to get tabs request, wait for network online target before booting decky

suchmememanyskill:
Tested with the help of @nabel0

Not 100% sure on the `network-online.target`. I think this only works when any network interface is up. With wifi disabled, after a reboot, decky still boots. I seem to have a loopback network interface that goes up pretty quickly, so that may be why. Someone who knows sytemd better than me should chime in on this


AAGaming00 (approved):
This will require users to reinstall devky to take effect, but LGTM

PR #240, opened 2022-10-30
Systemd service updating

TrainDoctor:
With @suchmememanyskill's recent updates to help with a potential race condition between plugins and loader, users would have to re-install decky to get their systemd services updated but now updates with update the service file as well.

This has been pretty been well tested but review is still requested.

### DIFF
--- a/backend/updater.py
+++ b/backend/updater.py
@@ -1,3 +1,5 @@
+import os
+import shutil
 import uuid
 from asyncio import sleep
 from ensurepip import version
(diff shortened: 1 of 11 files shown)


backend/updater.py
```diff
@@ -149,9 +195,10 @@ async def do_update(self):
                     out.write(version)
 
                 call(['chmod', '+x', path.join(getcwd(), "PluginLoader")])
-
+
```
AAGaming00, line 198:
```suggestion
```

AAGaming00 (changes requested):

AAGaming00 (approved):

PR #253, opened 2022-11-14
Fix desktop mode loop

### LINKED ISSUE #245: Switching from desktop to gaming mode does not always load decky
Hello!

I've noticed that when switching from desktop back to gaming mode decky does not always load, the icon will not show up. This appears to be random, because sometimes decky loads up just fine.

A possible workaround for this is to give the plugin_loader service a restart before going back to gaming mode.
I've been testing this for a couple days now and so far decky has loaded up every time when switching back and forth.

SteamOS stable update channel
Decky version v2.3.1-pre3
Installed plugins: PowerTools, SteamlessTimes

TrainDoctor:
Closes #245 when released via pre-release.

### DIFF
--- a/backend/injector.py
+++ b/backend/injector.py
@@ -36,8 +36,8 @@ class Tab:
 
     async def listen_for_message(self):
         async for message in self.websocket:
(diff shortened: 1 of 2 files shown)


TrainDoctor:
Ready for pre-release testing.

PR #258, opened 2022-11-18
Fix for wrong path for settings json files

RodoMa92:
With this, the files are copied automatically in the settings folder if present in the root directory (I'm scanning for file extension name (json), so if you will think to add json files to the root directory this will need to be changed) and copy them back to the settings folder. Permission issues on the settings folder are automatically fixed, both on creation and even if they are not matching after.

As far as my small tests are gone, it does seems to work perfectly fine now.

Let me know if I need to change anything else.

Marco

### DIFF
--- a/backend/helpers.py
+++ b/backend/helpers.py
@@ -1,3 +1,5 @@
+import grp
+import pwd
 import re
 import ssl
 import subprocess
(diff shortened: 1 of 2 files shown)


jurassicplayer:
Possibly chowning the .json, though it might not even be needed (only really benefits the handful of users that would change the config manually)

RodoMa92:
> Possibly chowning the .json, though it might not even be needed (only really benefits the handful of users that would change the config manually)

Whooops, you're right. Fixed it now.

RodoMa92:
I've moved two methods for getting the user and the group to a specific file in the helper module, in order to keep settings more readable and in case someone else would need that functionality. Unfortunately I would not know how to handle the case for each file created, and the only workaround that comes to mind is to try and use [https://docs.python.org/3/library/os.html?highlight=listdir#os.scandir](scandir) and check for the user owner for each file and change it to the user permission on load, or add a function to allow plugins that rely on the user to manually change the settings to call that and force the change of permission.

If anyone smarter than me have any suggestions, be my guest.

Marco

TrainDoctor:
Tested on my own Deck and seems to be working well. @AAGaming00 will be in charge of the review process for this PR.

AAGaming00 (approved):

PR #259, opened 2022-11-18
fixes: typos

MidSpike:


TrainDoctor:
Thank you for the PR but we don't generally accept two line change PRs. We will fix the typos regardless.

PR #268, opened 2022-12-02
feat(MoreCustomizableToasts): Allow plugin developers to customize some toast properties

jurassicplayer:
This PR does two main things:
- Replaced the settingsModule from Toaster with window.settingsStore because:
    - window.NotificationStore is already being used
    - from personal testing (SteamOS Preview, Decky-Loader Prerelease v2.4.4-pre1), decky-loader currently doesn't respect user-selected settings DisableAllToasts or DisableToastsInGame, and will send a toast when plugin updates are available
- Added properties matching those from SteamOS for `showToast`, `playSound`, `sound`, and `eType` to allow plugin developers the ability to have custom notification settings for their plugin (Toast+Sound, Toast only, Sound only, No notification)
    - showToast: boolean, shows a toast, defaults to true
    - playSound: boolean, plays sound, defaults to true
    - sound: number (0-25), audio sfx played when playSound is true, defaults to 6 (ToastMisc)
    - eType: number (1-24), type of toast, defaults to 11 as it seems to be closest in terms of properties to the current 15, but doesn't have a sfx tied to it

eType may be largely un-needed as the other three properties should cover most needs, but is included in-case another developer wants to explore some of the other toast types (of which many are somehow tied in with other features of SteamOS)

sound number to sfx mapping:
image

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -41,7 +41,7 @@
     }
   },
   "dependencies": {
-    "decky-frontend-lib": "^3.7.14",
+    "decky-frontend-lib": "^3.14.0",
     "react-file-icon": "^1.2.0",
     "react-icons": "^4.4.0",
     "react-markdown": "^8.0.3",
--- a/frontend/pnpm-lock.yaml
+++ b/frontend/pnpm-lock.yaml
@@ -10,7 +10,7 @@ specifiers:
   '@types/react-file-icon': ^1.0.1
   '@types/react-router': 5.1.18
   '@types/webpack': ^5.28.0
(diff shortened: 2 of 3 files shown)


TrainDoctor:
Please rebase your PR to latest changes to main so that when the CI runs decky can actually be built. (3.10.2 did not have a version provided by github in the actions cache so we've updated for now).
Also, we're aiming to review the DFL PR that introduces the needed changes for this PR hopefully this Friday. Thanks for your patience.

AAGaming00 (approved):
Needs a rebase and a DFL version bump, then ready to merge.

PR #269, opened 2022-12-07
Add custom CDN support for custom stores

NGnius:
Currently, custom stores are supported but the main CDN is always used. This prevents independent custom stores from existing.

This adds support for a providing a custom artifact url for every plugin version in the store. When that field is not provided (or null), it falls back to the main store's CDN.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,10 +47,10 @@ jobs:
       with:
         node-version: 18
       
-    - name: Set up Python 3.10.2 🐍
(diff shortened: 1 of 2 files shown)


TrainDoctor:
Please rebase your PR to latest changes to main so that when the CI runs decky can actually be built. (3.10.2 did not have a version provided by github in the actions cache so we've updated for now)

TrainDoctor (approved):
Tested locally on my own device, doesn't cause any compatibility issues from what I can tell. Merging to main and putting out a pre-release for me extensive testing.

PR #297, opened 2022-12-28
User Friendlier Installer

### LINKED ISSUE #193: Easier install/uninstall via a .desktop or .sh file (similar to EmuDeck)
This seems like a no-brainer, and i've seen it tossed around as an idea by many people.
After some thinking, i've realised there's a few questions that need to be thought about before it's done:

1. is it better to make one file that asks the user which version to install or is it better to have several files (one for each branch) or is it better to have one file that only installs stable, and make changing branches mainly done from within Decky
2. should uninstalling be part of the installer or a seperate file or part of decky as suggested in #147 (more than one of these is a perfectly reasonable option too)
3. what's the best way to format the file: is it better to do it like EmuDeck where the .desktop file just runs curl and the script is just grabbed off the internet or should the code be included with the downloaded file? (if the script(s) are included it'll work offline but be more complex, and possibly better suited as a .sh file instead of a .desktop)
4. what's a suitable level of complexity: is it worth putting a lot of extra work in for progress bars etc or should it be kept simple

i'm sure there's other things that need to be considered too.

PartyWumpus:
Adds a new installer (leave old ones to not break old guides, and also because emudeck uses it and it's faster etc) which is all UI based, optionally works without a password set and automatically enables CEF debugging.

The .desktop file works by downloading the script and putting it in /tmp/ and then running it from there. This is both because the code was easier to write if the file was actually stored on the system somewhere, and so it can (at some other time) be signed and checked relatively easily.

Tested on my my steam deck (with and without a password set) and my endevourOS machine and it appears to work properly.
Once it's verified it's working properly and any appropriate changes are made to the installer, the readme should probably also be changed to reflect this install method.

Would finally resolve #193, sorry I didn't get round to this sooner.

some screenshots:
image
(uninstall is only shown if decky appears to be installed)
image
image
image
image


PartyWumpus:
The installer should probably set the selected branch in settings too I suppose?

PartyWumpus:
Nevermind, looks like decky can already guess what branch is selected based on the version name?

AAGaming00:
> Nevermind, looks like decky can already guess what branch is selected based on the version name?

Correct. It will do this only if the value is unset in settings, though.

PartyWumpus:
so people can actually test it: [a version of the .desktop file that points at my repo](https://github.com/SteamDeckHomebrew/decky-loader/files/10316144/decky.txt) (you'll need to rename it to `decky.desktop` as github doesn't allow uploading of .desktop files)

dist/user_install_script.sh
```diff
@@ -0,0 +1,163 @@
...
+    rm -rf "/tmp/plugin_loader"
+    rm -rf "/tmp/user_install_script.sh"
+
+    echo "100" ; echo "# Cleaning services folder" ;
+    sudo rm "${HOMEBREW_FOLDER}/services/PluginLoader"
+    ) |
+    zenity --progress \
+  --title="Decky Installer" \
+   --width=300 --height=100 \
+  --text="Installing..." \
```
AAGaming00, line 80: Shouldn't this say uninstalling

dist/user_install_script.sh
```diff
@@ -0,0 +1,163 @@
...
+        if ( zenity --title="Decky Installer" --width=300 --height=200 --question --text="You appear to have not set an admin password.\nDecky can still install by temporarily setting your password to 'Decky!' and continuing, then removing it when the installer finishes\nAre you okay with that?" ); then
+            yes "Decky!" | passwd deck
+            trap temp_pass_cleanup EXIT # make sure password is removed when application closes
+            PASS="Decky!"
+        else exit 1; fi
+    else
+        # get password
+        FINISHED="false"
+        while [ "$FINISHED" != "true" ]; do
+            PASS=$(zenity --title="Decky Installer" --width=300 --height=100 --entry --hide-text --text="Enter your admin password")
```
AAGaming00, line 25: I'd assume more people would know this as a "sudo password" rather than "admin password" as a lot of guides refer to it as the sudo password iirc

AAGaming00 (changes requested):
See comments

dist/user_install_script.sh
jurassicplayer: Or just `sudo/admin` to cover both

dist/user_install_script.sh
PartyWumpus: Yep, although it will probably never actually show up as the first command run replaces it

dist/user_install_script.sh
PartyWumpus: did `sudo/admin`

PartyWumpus:
thanks to @Godsbane, there's now a downloading bar when it's getting the pluginLoader file
image

PartyWumpus:
Is it worth adding a 'manual install' option that lets users grab a PluginLoader file from their filesytem?
I'm wondering if it's possible to make it work better for those without internet on their deck (due to region or whatever) although they'll probably be using the script directly instead of through the .desktop file.
Shouldn't be too hard to implement but not sure if it's worth it

Something like this that opens a file manager prompt (with `zenity --file-selection`)
image


Other than this and eventually adding signature validation stuff (I can't do this now obviously as the decky releases + the script aren't signed yet) the installer is probably _actually_ done now (beyond any suggested changes).

TrainDoctor:
This is perfect as is. Merging to main.

PR #299, opened 2022-12-28
Add cef debugging to the installer scripts

PartyWumpus:
Adds automatic enabling of cef debugging in the (soon to be old) installers.
Works fine on my deck.

Might be worth adding an `echo "Enabling CEF debugging"` to the file to alert the user, but probably not.


PartyWumpus:
Need to close and reopen to change the branch it's coming from

Issue #280, opened 2022-12-13
Improve store design


EMERALD0874:
- Improve the store design to be more modern but still different from Valve's design. I will use the Audio Loader and CSS Loader plugins as inspiration for this project.
- Add sorting and searching to the store. This will be taken from the Audio Loader under the MIT License.
![image](https://user-images.githubusercontent.com/11338953/207209366-54476430-8c7d-4e2b-bdf2-950f6ca1b78a.png)

PR #305, opened 2022-12-30
Updated store CSS

GoDev87:


TrainDoctor:
@GoDev87 could you provide some before and after screenshots for reference? Thanks.

GoDev87:
I have class right now, I'll do that in ~8h.

GoDev87:
image
@TrainDoctor

GoDev87:
image
And this is without any css plugins

EMERALD0874 (commented):
This will only be briefly helpful as the whole store will be redesigned in #280 and deviate from the current structure. I'm not sure if it's worth merging these changes.

TrainDoctor:
> This will only be briefly helpful as the whole store will be redesigned in #280 and deviate from the current structure. I'm not sure if it's worth merging these changes.

A decent point but it won't hurt to merge it as a qol measure when we make the next stable release. Merging to main.

PR #318, opened 2023-01-06
Add feature requests as an issue template

PartyWumpus:
adds feature requests as another issue type. 
(if you don't want to merge this, then `.github/ISSUE_TEMPLATE/config.yml` should probably have `blank_issues_enabled` re-enabled or feature requests will have to fit within the bug request template which is annoying)

image

----
image


TrainDoctor (approved):
Looks great, better to optimize the info flow. Cheers.

PR #324, opened 2023-01-08
Use the new installer in the readme instructions 

### LINKED ISSUE #321: [Request] Use the new installer in the readme instructions
### Please confirm

- [X] This is not a request for a plugin

### Feature Request Description

The new GUI installer should probably be listed in the readme because it'll be easier for users etc (also otherwise I wasted a weekend making it :p)
The fewer steps in the download process, the better. If possible, a big download button that sends them straight to the .desktop that they can then just run and have work would be amazing. (also, the old installers should probably still be mentioned somewhere in the readme for users who want a fast install)

### Further Description

There seems to be a few ways of having a download button (and they're all not great imo):

1. Download button that sends users to a github pages site (or a different site i guess) that links the user to the most recent version of the file. An example that uses this is [Viper](https://github.com/0neGal/viper). This, and the second option, have to have the .desktop file in releases as github doesn't seem to let you download other files via a browser.

2. Download button sends that users directly to the most recent version of the file  and [this action](https://github.com/DenverCoder1/readme-download-button-action) will update the readme every time there's a release

3. Download button uses [DownGit](https://minhaskamal.github.io/DownGit/) with a [link like this](https://minhaskamal.github.io/DownGit/#/home?url=https://github.com/SteamDeckHomebrew/decky-loader/blob/main/dist/decky_installer.desktop) to download it. I don't love this idea, but anyway it appears it downloads every file as a .zip (for now) so isn't viable.

Without a button:

1. Put the installer in releases and just ask people to go there and use it

2. Ask people to run `curl -L ... | sh` like before. This is not a good solution and is also not really compatible with adding signed releases later... Also the main goal of the installer is stopping people from having to use the terminal, so that would somewhat defeat the point.

I'm sure there's some other options (probably something obvious), but I just wanted to make this issue to actually get the ball rolling on using the new installer. With all these options I'm not going to start making a PR for it without some input. (But I am happy to do it myself if possible)

PartyWumpus:
Resolves #321

The .desktop file is now included in Releases as that seems to be the best way to handle easy downloads (readme download button just links to `https://github.com/SteamDeckHomebrew/decky-loader/releases/latest/download/decky_installer.desktop`). Because of this change I also had to make a small modification to the installer scripts so they grab specifically the PluginLoader file instead of just everything in releases.
image

------

Readme now looks like this:
image
image

-----

The readme instructions have been changed, but I'm not very good at writing instructions so they should probably be rewritten at some point. Also not sure about the looks of the download button, I just wanted something so I made that quickly in krita. If someone could get a better button that would be nice.


MortalAngel:
download buttons dont works.. yesterday i was able to download the installer..today not when i tried on the second deck..same on pc

PartyWumpus:
@MortalAngel  yeah this is a known issue and should hopefully be resolved soon. For now you can manually download this file and call it 'decky.desktop' https://github.com/SteamDeckHomebrew/decky-loader/raw/main/dist/decky_installer.desktop

PartyWumpus:
@MortalAngel  yeah this is a known issue and should hopefully be resolved soon. For now you can manually download this file and call it 'decky.desktop' https://github.com/SteamDeckHomebrew/decky-loader/raw/main/dist/decky_installer.desktop

PR #327, opened 2023-01-09
Make updater work when there is more than one file in releases

PartyWumpus:
Hopefully fixes the updater by looking for the PluginLoader asset instead of just the first one.

This was done on my phone so is likely to be wrong.


PartyWumpus:
This is no longer neccecary, but appears to work and could be nice to have when releases are signed. Will do further testing just in case.

PartyWumpus:
Hmm. It's freezing up the entire deck when ever I try and update, but this is also happening when I try and use main instead. Just don't merge this for now.

TrainDoctor:
PR is stale. Closing unless developer wishes to return to concept.

PR #347, opened 2023-01-20
Inject only in gamepad ui mode

popsUlfr:
Additional safeguard using 'SteamClient.UI.GetUIMode()' to only inject decky in gamepad ui mode (or gaming mode). Injecting decky into the desktop mode Steam client could lead to unintended behaviour.

Some users using plugins like AudioLoader have reported that they would seemingly be active even with the desktop mode Steam client. Plugin developers usually don't account for that possibility and it could lead to strange and unintended behaviour.

`SteamClient.UI.GetUIMode()` seems to do the trick to determine in which UI mode Steam is running. It returns `0` for desktop mode Steam Client and `4` for gamepad ui mode Steam Client.

I'm having trouble reproducing the issue on my end unfortunately, decky does not even manage to inject itself into the Steam Client for me. I've tested the inverted condition though with `(await window.SteamClient.UI.GetUIMode()) !== 4)` and it successfully prevents decky from injecting itself in gamepad ui mode.
The themes from CSSLoader still seem to get injected though, so there might be some more logic needed to completely prevent CSS injection too but it might not be needed if the execution of the plugin code itself is avoided.

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -172,7 +172,7 @@ class PluginManager:
             if first:
                 if await tab.has_global_var("deckyHasLoaded", False):
                     await close_old_tabs()
-            await tab.evaluate_js("try{if (window.deckyHasLoaded){setTimeout(() => location.reload(), 100)}else{window.deckyHasLoaded = true;(async()=>{try{while(!window.SP_REACT){await new Promise(r => setTimeout(r, 10))};await import('http://localhost:1337/frontend/index.js')}catch(e){console.error(e)};})();}}catch(e){console.error(e)}", False, False, False)
+            await tab.evaluate_js("try{if (window.deckyHasLoaded){setTimeout(() => location.reload(), 100)}else{window.deckyHasLoaded = true;(async()=>{try{while(!window.SP_REACT){await new Promise(r => setTimeout(r, 10))};if (!window?.SteamClient?.UI?.GetUIMode || (await window.SteamClient.UI.GetUIMode()) === 4) await import('http://localhost:1337/frontend/index.js')}catch(e){console.error(e)};})();}}catch(e){console.error(e)}", False, False, False)
         except:
             logger.info("Failed to inject JavaScript into tab\n" + format_exc())
             pass


CarJem:
I demand that this be considered something plugins can decide (But off by default)

I'm working on a plug-in called DeckyBoard That will be able to add additional keys to the keyboard... The opportunity should still be possible to patch outside of game mode!

popsUlfr:
> I demand that this be considered something plugins can decide (But off by default)
> 
> I'm working on a plug-in called DeckyBoard That will be able to add additional keys to the keyboard... The opportunity should still be possible to patch outside of game mode!

I understand. The bigger problem right now is that I personally haven't been able to get decky inject itself into the desktop Steam Client so from what I see decky patching outside of gaming mode is akin to a glitch. This is more of a temporary safeguard against something like this happening because there's potential for weird issues to manifest if the plugin author didn't account for that scenario.

EMERALD0874:
Decky should still be loaded, but individual plugins should not be injected unless they have a boolean set to true in a plugin.json. Most plugins (ex. CSS Loader, Audio Loader) have no reason to be injected, but others could see some use.

I'm afraid I have to disagree with merging this PR until we see a solution that doesn't limit future options.

popsUlfr:
The reality right now is that plugins like AudioLoader where a soundpack can be selected and played in the background and controlled through decky in the quick access menu, may also start playing in the Desktop Steam Client version where there's no quick access menu and thus no way for the user to have any control over the plugins. As a user I would consider this bug.

Now more so, decky injecting itself into the Desktop Steam Client feels like an unreliable accident since it does not seem to work for everyone. This should in my opinion be improved as part of feature addition to decky that aims to add proper desktop Steam Client support to decky.

But currently how I see it, is that this is simply a little safeguard against an unreliable and unintended behaviour (decky potentially injecting itself into the desktop Steam Client) that never had proper support in the first place.

Therefore the proper support for desktop Steam Client should be contributed as part of a feature request where users can properly control their plugins even in the desktop Steam Client. (and to be extra clear, decky will still inject itself into the new Big Picture mode (aka gamepad ui like in gaming mode) since it's the same UI mode as in gaming mode. This is jsut about the desktop Steam Client in normal mode)

PR #349, opened 2023-01-20
Expose useful env vars to plugin processes

popsUlfr:
This change aims to suggest recommended settings, runtime data, logs paths to plugins through environment variables in order to propose a common storage guideline for plugins to follow.

In doing so I've cleaned up and hopefully improved the helper functions for the paths a bit. For instance the user is deduced from the path the `PluginLoader` is launched from by matching with the home directory specified in the passwd entry.

The environment variables that are exposed and can be retrieved from a python script with `os.environ["DECKY_HOME"]` or the appropriate functions when using a custom backend binary are as follows:

- `HOME`: the home directory of the effective user running the process
  + e.g.: if `root` was specified in the plugin's flags it will be `/root` otherwise the user whose home decky resides in
- `USER`: the effective username running the process
  + e.g.: it would be `root` if `root` was specified in the plugin's flags otherwise the user whose home decky resides in
- `DECKY_VERSION`: the version of the decky loader
  + e.g.: `v2.5.0-pre1`
- `DECKY_USER`: the user whose home decky resides in
  + e.g.: `deck`
- `DECKY_HOME`: the root of the decky folder
  + e.g.: `/home/deck/homebrew`
- `DECKY_PLUGIN_SETTINGS_DIR`: the recommended path in which to store configuration files (created automatically)
  + e.g.: `/home/deck/homebrew/settings/decky-plugin-template`
- `DECKY_PLUGIN_RUNTIME_DIR`: the recommended path in which to store runtime data (created automatically)
  + e.g.: `/home/deck/homebrew/data/decky-plugin-template`
- `DECKY_PLUGIN_LOG_DIR`: the recommended path in which to store persistent logs (created automatically)
  + e.g.: `/home/deck/homebrew/logs/decky-plugin-template`
- `DECKY_PLUGIN_DIR`: the root of the plugin's directory
  + e.g.: `/home/deck/homebrew/plugins/decky-plugin-template`
- `DECKY_PLUGIN_NAME`: the name of the plugin as specified in the `plugin.json`
  + e.g.: `Example Plugin`
- `DECKY_PLUGIN_VERSION`: the version of the plugin as specified in the `package.json`
  + e.g.: `0.0.1`
- `DECKY_PLUGIN_AUTHOR`: the author of the plugin as specified in the `plugin.json`
  + e.g.: `John Doe`

Tell me if you find these adequate enough for helping plugins stay tidy with their paths.

I've kept `set_user` and `set_user_group` around as deprecated functions since some plugins call them and would crash without these.

The next step would be to update the plugin template to acknowledge these variables and provide a starting point or update on the best practices when storing data.

### DIFF
--- a/backend/helpers.py
+++ b/backend/helpers.py
@@ -5,6 +5,7 @@ import ssl
 import subprocess
 import uuid
 import os
+import sys
 from subprocess import check_output
 from time import sleep
 from hashlib import sha256
@@ -19,8 +20,6 @@ REMOTE_DEBUGGER_UNIT = "steam-web-debug-portforward.service"
 # global vars
 csrf_token = str(uuid.uuid4())
 ssl_ctx = ssl.create_default_context(cafile=certifi.where())
-user = None
-group = None
 
 assets_regex = re.compile("^/plugins/.*/assets/.*")
 frontend_regex = re.compile("^/frontend/.*")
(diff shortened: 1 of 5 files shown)


TrainDoctor (approved):
I've added some documentation so plugin developers will have references to these variables existing. Let's take this for a spin!

PR #356, opened 2023-01-31
Add support for more hardware when building

EMERALD0874:
By specifying the `linux/amd64` Docker container architecture, hardware that doesn't use `amd64` (ex. Apple silicon processors) is able to build Decky Loader artifacts using the `act/run-act.sh` script.

Please feel free to test with and without the provided changes and report the changes across different systems.

### DIFF
--- a/act/run-act.sh
+++ b/act/run-act.sh
@@ -26,10 +26,10 @@ cd ..
 
 if [[ "$type" == "release" ]]; then
     printf "release!\n"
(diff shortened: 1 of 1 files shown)


EMERALD0874:
**MacBook Pro (14", 2021) w/ M1 Processor**

- Before - Fails to install Python as `arm64` version does not exist
- After - Works as intended

PR #357, opened 2023-02-01
Fix UI inconsistencies, various improvements

### LINKED ISSUE #346: [BUG] UI inconsistencies and improvements
### Please confirm

- [X] I have checked the [common issues section in the readme file](https://github.com/SteamDeckHomebrew/decky-loader#-common-issues)

### Bug Report Description

Please note that I am only listing these inconsistencies so we can start a discussion on what should be fixed (if anything). I am being intentionally extremely detailed and understand not everything will be changed.

- [x] No other Quick Access tab has buttons in the header like this, creating a jarring difference in headers. It may be better to group the store and settings buttons on the right side.

image

- [x] 1. "DECKY SETTINGS" is not necessary as the context is understood by the surrounding options and previously pressed button. Valve does not use such a header in its options menu.
- [x] 2. Sections should have icons before them (ex. cog for Decky Loader, store for Plugins, whatever odd symbol Valve uses for Developer) and "General" should be changed to "Decky Loader" so it's more specific, as the settings there do not affect plugins.
- [x] 3. An orange exclamation mark icon is the only one shown in this position and only when a new update is available. This icon is unnecessary. The current version could also be moved to the bottom with the text instead displaying "Up to date: last checked X minutes ago" similar to how Valve does it. This prevents crowding (if an update is unavailable, of course the current and latest are the same) and may help as a sanity check if someone is worried they aren't getting updates.
- [x] 4. Icons are only used in Valve's settings if 100% descriptive of what is being changed or necessary to convey information (ex. Bluetooth, game rumble (a moving Steam Deck), Nintendo Button Layout (A and B buttons swapping positions). This should be removed as it could be confusing if a non-English speaker were browsing the menus. Also, "Allow Remote CEF Debugging" is not a proper noun and should not use title case. _Also_, it should probably be considered a developer setting. Possibly the same for manual plugin installs as most users should not be using stores outside of those provided by Decky (could lead to malicious installations).
- [x] 5. Buttons and dropdowns are wider in Steam Deck settings. Just a nitpick.
- [x] 6. Patch notes should only be visible if an update is available (like in Valve's implementation). Users can always press Y on the "Check For Updates" button if they need to view patch notes.

image

- [x] Settings within a larger menu should be grouped into smaller groups (ex. Updates should have its own Updates category, Update Channel and Store Channel could be a part of Beta Participation, etc.)

image

- [x] Maybe put the version in parentheses and darker text? Similar to how Steam does friend nicknames. The version is secondary information and should not be a part of the plugin name itself.

image

### Expected Behaviour

The expected behavior is described alongside individual inconsistencies for easier reading. Checkboxes are provided for developers working on the ticket.

### SteamOS version

Stable 3.4.4

### Selected Update Channel

Stable

### Have you modified the read-only filesystem at any point?

Nope!

### Logs

_No response_

EMERALD0874:
### Quick Access Menu

- Groups the settings and store buttons on the right so the header lines up with other quick-access menu headers.
- Moves the store button left of the settings button as installing plugins is the primary purpose of Decky Loader.

image

### Settings - Navigation

- Removes the Decky Settings header at the top of the navigation, as users know where they navigated to and Valve does not use a header for the Deck settings.
- Changes "General" to "Decky" so it's clear the settings are for Decky in particular.

### Settings - All Pages

- Puts all settings inside of DialogBody to assist with scrolling behavior.

### Settings - Decky

- Adds section headers and uses DialogControlsSection to assist navigation.
- Clarifies "Updates" are "Decky Updates".
- Makes current/latest version text only show text (Up to date: running [version]) if there is no newer version.
- Hides patch notes dialog unless there is a new version available.
- Changes "Update Channel" and "Store Channel" to "Decky Update Channel" and "Plugin Store Channel" respectively to be clearer.
- Removes unnecessary icons and descriptions.
- Adds an About section that contains the current Decky Loader version (moved from the top, more similar to Valve's design philosophy).

image

### Settings - Plugins

- Applies 50% opacity to plugin versions and put them in parenthesis (visual distinction of version from plugin name).
- Removes the left padding from the list of plugins.
- Uses DialogControlsSection to assist in scrolling to the top.

image

### Settings - Developer

- Moves Allow Remote CEF Debugging to the Developer settings as it presents a risk for end users and is only used in development and debugging.
- Adds extra spacing for Enable React DevTools URL field.
- Changes the Enable React DevTools icon to Chrome as the debugging icon could be used for other settings and isn't very clear.
- Fixes Enable React DevTools functionality without the touchscreen.
- Uses Valve's provided attribute for determining Developer page visibility instead of adding to an array based on a boolean.

image

### Other

- Adds a React Icon-like DeckyIcon object.
- Closes #346.


TrainDoctor (approved):
Looks good, let's give it some pre-release testing.

PR #361, opened 2023-02-02
[Feature] Implement internazionalization for Decky Loader

RodoMa92:
As promised, this is my prototype for translating this awesome piece of software.

For translating the codebase, I've chosen i18next as a framework.

Whats seems to work fine:

1) Backend serves json files corresponding to the Steam user language selected plus a fallback to english by default
2) Frontend had all strings replaced with the corresponding t() call (see below for details) in order to dynamically update the text in the UI.
3) English and Italian complete translation are already provided, although I will probably tweak it again as soon as I begin to test a lot more of the UI
4) Strings are automatically extracted using the action extracttext from vscode. This will scan the codebase, search for t() calls and fill automatically the json translation files with empty values (can be set to whatever we want).
5) AFAICS settings UI seems already to works perfectly fine.

Possible pitfalls left in this:
1)  I have seen a lot of https://reactjs.org/docs/error-decoder.html/?invariant=321 in the console of chromium while debugging the backend, I might still have left some boobytraps around the codebase, I still need to actually check if they are still there and if they actually are an issue. UPDATE: This is caused by some components missing react hooks inside them (I think), for now I should have disabled the small translation stuff caused by this, and it's working fine now (but not translated). If someone wants to take a stab at it at fixing them, I left the translated part commented out here.
2) The whole codebase still needs to be heavily tested.

Information to the devs in order to translate text:
1) First you need to import the main function from i18next:
`import { useTranslation } from 'react-i18next';`

2) After that you need to create a hook inside the corresponding react function like this:
`const { t } = useTranslation();`

3) This will give you the function t that you can use inside react components in order to resolve text like this:
`t('ClassName.corresponding_string')`. The first parameter is the identifier for resolving the desired text. 

i18next have quite a lot of options for adapting to various languages, that can be specified after the string identifier. I recommend translators to go read the documentation, it's quite succinct and complete in that regard.

4) Once all parameters have been added to the file, run extracttext. This will parse the codebase and fill up all json files automatically.

If you want to add support for an additional missing language to the codebase, the steps are the following:

1) Add your language identifier (ex: en-US for american english, it-IT for italian, etc.) in the array at row 47 in the file called `i18next-parser.config.mjs`.

2) Run the activity `extracttext` in visual studio code. This will parse automatically the codebase and generate a corresponding json file in `<root_codebase>/backend/locales` for your languages, all with parameters set to `MISSING` (also configurable on demand).

3) ~~If you can't translate everything, remember to remove the parts of the JSON not translated, in order to allow i18next to fallback to the default english language. Otherwise the UI in that specific language will look full of `MISSING` everywhere. This is the reason I've dropped the empty language files in the latest commits, to avoid any confusion. If anyone wants to translate it further, you just need to run `extracttext` again to get all the non translated strings back in the file.~~ Not necessary anymore.


david082321:
I think you can add support for subtypes of language types. 
Because Chinese had many type, and they had a big different. 
e.g. zh-CN (zh-Hans) and zh-TW (zh-Hant).

RodoMa92:
> I think you can add support for subtypes of language types. Because Chinese had many type, and they had a big different. e.g. zh-CN (zh-Hans) and zh-TW (zh-Hant).

Working on it now, should be relatively easy to change.

RodoMa92:
It's in, now it's supported.

suchmememanyskill (changes requested):
Reviewing the backend part of this PR: I suspect the `locales` folder should be part of the binary, but such changes are not reflected in `/.github/workflows/build.yml`, [line 72](https://github.com/SteamDeckHomebrew/decky-loader/blob/0dd0d9f4bd409852a8a7eab88ecf07b608115d72/.github/workflows/build.yml#LL72C4-L72C4). I suspect adding `--add-data ./locales:/locales` would do the trick, but please do test that.

RodoMa92:
> Reviewing the backend part of this PR: I suspect the `locales` folder should be part of the binary, but such changes are not reflected in `/.github/workflows/build.yml`, [line 72](https://github.com/SteamDeckHomebrew/decky-loader/blob/0dd0d9f4bd409852a8a7eab88ecf07b608115d72/.github/workflows/build.yml#LL72C4-L72C4). I suspect adding `--add-data ./locales:/locales` would do the trick, but please do test that.

Looking at the same line, it's probably should be `--add-data ./backend/locales:locales`, added it now. Thanks for the catch, I didn't see that.

TrainDoctor (approved):
This just needs to testing by users, otherwise I believe this is good to merge once it has received testing.

Witherking25:
Y'all realize valve has localization built into steam ui

Witherking25:
I've seen it in the components, the actual text looks like a localized string because is has like a # at the start and what actually gets displayed looks like it is indexed from a table depending on the current language and the localization tag (the bit of text that valve actually typed into the component)

Witherking25:
I was going to figure out how to utilize it in my plugins, so maybe it is worth looking into for decky

TrainDoctor:
> Y'all realize valve has localization built into steam ui

That is an entirely Valve thing using a process similar to this... Valve does not auto-translate text in components.

Witherking25:
I'm pretty sure they do, I'll find an example after class

RodoMa92:
> Y'all realize valve has localization built into steam ui

Valve for sure uses it's own framework for translate text in their UI. However this patch do not use anything provided by them, since we do not really need much besides all hooks needed to interface and patch their own web interface and inject our code.

Regardless, since we do not have the actual source code of Big Picture from Valve what they uses doesn't really matters much to us. This already integrates with the Steam language selector, so it will behave exactly as the Steam UI components anyway, without any difference to the user.

RodoMa92:
> I was going to figure out how to utilize it in my plugins, so maybe it is worth looking into for decky

This is planned as an API down the line IIRC, but not for now. @AAGaming00 will know better than me. This exclusively support translation for the Decky Web Interface and nothing else. The API to fetch the corresponding translated text from the backend is not included in this PR.

Witherking25:
Also, not everyone uses vscode, so making it a script that could be called by whatever ide you are using would be preferred. For example I use intellij idea ultimate

RodoMa92:
> Also, not everyone uses vscode, so making it a script that could be called by whatever ide you are using would be preferred. For example I use intellij idea ultimate

You can get the corresponding commands from the tasks.json, first [this command](https://github.com/RodoMa92/decky-loader/blob/0a789807e14df6a3961fa22667bd5a473020d4aa/.vscode/tasks.json#L61`) and then [this command](https://github.com/RodoMa92/decky-loader/blob/0a789807e14df6a3961fa22667bd5a473020d4aa/.vscode/tasks.json#L52). Feel free to wrap your own script to do this.

TrainDoctor:
@SteamDeckHomebrew/decky-testers please give this a try and report any issues.

RodoMa92:
If any tester want to try the code, download a snapshot from 0a789807e14df6a3961fa22667bd5a473020d4aa, the master is messed up until this evening @CEST until I get the time to fix up some merge issues. Sorry for the confusion.

RodoMa92:
Now it has been updated, you can test the latest commits.

TrainDoctor:
Sorry that we haven't touched this very much since I provided my approving review. Once the current reorderable issues are solved I plan with getting this tested and merged asap.

frontend/src/components/modals/TPluginInstallModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+import { InstallType } from '../../plugin';
+
+export enum TranslatedPart {
+  TITLE,
+  DESC,
+  BUTTON_IDLE,
+  BUTTON_PROC,
+}
+class TPluginInstallModal extends Component<TranslationProps> {
```
AAGaming00, line 12: Is there a way to do this without class components? We're trying to avoid using them as much as possible.

backend/main.py
```diff
@@ -94,6 +94,7 @@ async def startup(_):
         for route in list(self.web_app.router.routes()):
             self.cors.add(route)
         self.web_app.add_routes([static("/static", path.join(path.dirname(__file__), 'static'))])
+        self.web_app.add_routes([static("/locales", path.join(path.dirname(__file__), 'locales'))])
```
AAGaming00, line 97: This can probably be deleted as loader.py handles it.

frontend/src/components/modals/PluginInstallModal.tsx
```diff
@@ -1,18 +1,29 @@
...
+  artifact,
+  version,
+  hash,
+  installType,
+  onOK,
+  onCancel,
+  closeModal,
+}) => {
   const [loading, setLoading] = useState<boolean>(false);
+
```
AAGaming00, line 26: Did prettier add this or did you add this?

frontend/src/components/settings/pages/general/BranchSelect.tsx
```diff
@@ -14,17 +15,23 @@ enum UpdateBranch {
 }
 
 const BranchSelect: FunctionComponent<{}> = () => {
+  const { t } = useTranslation();
+  const tStores = [
+    t('StoreSelect.store_channel.default'),
+    t('StoreSelect.store_channel.testing'),
+    t('StoreSelect.store_channel.custom'),
+  ];
```
AAGaming00, line 23: This is wrong, it is for decky branches not store channels.

frontend/src/components/settings/pages/general/StoreSelect.tsx
```diff
@@ -11,17 +12,23 @@ const logger = new Logger('StoreSelect');
 const StoreSelect: FunctionComponent<{}> = () => {
   const [selectedStore, setSelectedStore] = useSetting<Store>('store', Store.Default);
   const [selectedStoreURL, setSelectedStoreURL] = useSetting<string | null>('store-url', null);
+  const { t } = useTranslation();
+  const tBranches = [
```
AAGaming00, line 16: Again, this is for store selection, not update channel.

frontend/src/components/settings/pages/plugin_list/index.tsx
```diff
@@ -25,19 +27,33 @@ async function reinstallPlugin(pluginName: string, currentVersion?: string) {
...
   const data = props.entry.data;
+  const { t } = useTranslation();
   let pluginName = labelToName(props.entry.label, data?.version);
 
   const showCtxMenu = (e: MouseEvent | GamepadEvent) => {
     showContextMenu(
-      <Menu label="Plugin Actions">
-        <MenuItem onSelected={() => window.DeckyPluginLoader.importPlugin(pluginName, data?.version)}>Reload</MenuItem>
-        <MenuItem onSelected={() => window.DeckyPluginLoader.uninstallPlugin(pluginName)}>Uninstall</MenuItem>
+      <Menu label={t('PluginListIndex.plug_actions_label')}>
```
AAGaming00, line 41: would plugin_actions be better?

frontend/src/index.tsx
```diff
@@ -36,9 +40,34 @@ declare global {
...
+    .init({
+      load: 'currentOnly',
+      detection: {
+        order: ['querystring', 'navigator'],
+        lookupQuerystring: 'lng',
+      },
+      //debug: true,
+      fallbackLng: 'en-US',
+      interpolation: {
+        escapeValue: false,
```
AAGaming00, line 56: what's this do?

frontend/src/index.tsx
```diff
@@ -36,9 +40,34 @@ declare global {
...
+      },
+      backend: {
+        loadPath: 'http://127.0.0.1:1337/locales/{{lng}}.json',
+        customHeaders: {
+          Authentication: window.deckyAuthToken,
```
AAGaming00, line 61: you dont need this if you added it to the ignore list in helpers.py, choose to require auth or remove it from there

frontend/src/plugin-loader.tsx
```diff
@@ -294,12 +321,14 @@ class PluginLoader extends Logger {
...
+    const { t } = useTranslation();
+
     return new Promise((resolve, reject) => {
       const Content = ({ closeModal }: { closeModal?: () => void }) => (
         // Purposely outside of the FilePicker component as lazy-loaded ModalRoots don't focus correctly
         <ModalRoot
           onCancel={() => {
-            reject('User canceled');
+            reject(t('PluginLoader.file_picker_cancel_text'));
             closeModal?.();
```
AAGaming00, line 332: APIs like this that are passed to plugins should never be localized! This can cause issues in plugins and is never displayed to users!

frontend/i18next-parser.config.mjs
```diff
@@ -0,0 +1,102 @@
+// i18next-parser.config.js
```
AAGaming00, line 1: this can probably be deleted lol

AAGaming00 (changes requested):
See comments

frontend/src/components/modals/PluginInstallModal.tsx
RodoMa92: Prettier, this doesn't come from me :P

frontend/src/components/modals/TPluginInstallModal.tsx
RodoMa92: This is a previous iteration without any additional components, but it looked ugly to me. If it's fine for you, I can revert this to that: https://github.com/RodoMa92/decky-loader/blob/91b2a1b38545bffd5d684e06e5ec9a1a20ca468c/frontend/src/components/modals/PluginInstallModal.tsx

AAGaming00: The additional component is fine, but I'd prefer a functional component instead of a class component.

frontend/i18next-parser.config.mjs
RodoMa92: Agreed

frontend/src/plugin-loader.tsx
RodoMa92: Reverted now.

frontend/src/index.tsx
RodoMa92: How i18next should handle tag injection inside their properties, see here:

https://www.i18next.com/translation-function/interpolation#unescape

frontend/src/components/settings/pages/general/BranchSelect.tsx
RodoMa92: Whoops, fixing it now.

frontend/src/components/settings/pages/general/StoreSelect.tsx
RodoMa92: Also fixed this.

frontend/src/index.tsx
RodoMa92: Removed the bypass on loader.py

frontend/src/components/modals/TPluginInstallModal.tsx
RodoMa92: Fixed as of latest commit, replaced with a functional component.

RodoMa92:
I should have fixed everything, let me know if I have missed something.

AAGaming00:
I feel like there may be a better way to handle translation in the modals to allow for usage of useTranslation, which I'll look into before merging this

frontend/src/index.tsx
AAGaming00: IMO, leave escaping enabled as it will surely cause later headache for us leaving it off. For long, formatted documents like the store About page, it might be better to convert those to Markdown and use the MarkdownRenderer (and also move that component to not be specialized for the patch notes).

AAGaming00:
> If you can't translate everything, remember to remove the parts of the JSON not translated.

Is there a way to automate this?

RodoMa92:
> > If you can't translate everything, remember to remove the parts of the JSON not translated.
> 
> Is there a way to automate this?

In theory you can pass a function to default value that should allow to disable the output if needed.
`// You may also specify a function accepting the locale, namespace, key, and value as arguments`

frontend/src/index.tsx
RodoMa92: Reenabled the manual escape as requested.

RodoMa92:
> I feel like there may be a better way to handle translation in the modals to allow for usage of useTranslation, which I'll look into before merging this

A few pointers for you:
The only feasible way that I found to change the message programmatically in i18next is to use the count property. There are two ways of doing this:

- Native count: This is not deterministic, unfortunately (each language might have different ways to enumerate. Some gives a lot of them, other only two). It's not really useful in this case.
- Using a first party extension [link](https://github.com/i18next/i18next-intervalplural-postprocessor): This allows exactly what we want (we can specify a defined interval and the associating text for each language freely), however I couldn't get to cooperate with the parser, and the defined strings would be seen as not used and removed each time I would run the parser. It's probably possible to coax the parser by specifying additional options to it (probably in i18nextOptions), but I didn't explore this further personally. This is probably the easiest route.
- Maybe a third party postprocessor? Still, similar issue as above.

Regarding the toaster, I've tried to inject the property in it to no avail, but since this is a interface to plugins, and not only internally to Decky, I'll prefer to leave this to someone who knows the code better than me.

Otherwise this branch has been merged again with the latest master.

PartyWumpus:
This looks good to merge, I've been using this for a while without issue, and Train's been testing it over the weekend. Will merge if you think it's ready (and make a prerelease with it in).

RodoMa92:
Besides the two things above (the not ideal modal translation selection and the notifications not translated yet, but already integrated in the translation system) I do not see any other missing features or bugs left in this branch. In my opinion it's ready to merge.

PR #365, opened 2023-02-04
Add linting to Python files in CI

SkyLeite:
Should not  be merged before #363

Feel free to suggest and/or add configurations to both Black and Ruff in `backend/pyproject.toml` if anything stands out. In the process of fixing linter errors I ended up fixing a couple minor bugs as well (in code that doesn't seem to be called at least).


AAGaming00:
outdated

PR #378, opened 2023-02-22
Add plugin reordering

Tormak9970:
Update to the plugin list to allow users to reorder their plugins. Will save the install order as default the first time the loader runs and the order setting is missing.


frontend/src/components/settings/pages/plugin_list/index.tsx
```diff
@@ -2,24 +2,93 @@ import {
...
+          Reload
+        </MenuItem>
+        <MenuItem onSelected={() => window.DeckyPluginLoader.uninstallPlugin(props.entry.label)}>Uninstall</MenuItem>
+      </Menu>,
+      e.currentTarget ?? window,
+    );
+  };
+
+  return (
+    <Fragment>
```
AAGaming00, line 35: Do not use `Fragment`, use `<>`. The loader tsconfig is set up to allow `<>` to work.

.vscode/defsettings.json
```diff
@@ -1,7 +0,0 @@
-{
```
AAGaming00, line 1: Why was this file deleted?

frontend/pnpm-lock.yaml
```diff
@@ -11,7 +11,7 @@ specifiers:
   '@types/react-file-icon': ^1.0.1
```
AAGaming00, line 11: Please do not run a formatter on this file. Prettier should only be ran on the `src` directory.

AAGaming00 (dismissed):
See comments.

.vscode/defsettings.json
Tormak9970: don't recall removing it. will add it back

TrainDoctor:
Requesting review from @AAGaming00 since noted issues have been resolved.

TrainDoctor:
Needs a conflict resolution for ``browser.py``.

EMERALD0874 (commented):
LGTM, though the edits to [.vscode/defsettings.json](https://github.com/SteamDeckHomebrew/decky-loader/pull/378/files#diff-d134d3795c9070a0d30511a5e60aeb9038f462c1ad81b69decf0184d6b210357) are unnecessary. Just formatting changes though so it's fine imo

Tormak9970:
@TrainDoctor Should be fixed

TrainDoctor (approved):
Looks good to go, going to do some local testing, requesting @SteamDeckHomebrew/decky-testers give it a whirl as well but that will not be required for merge. Thanks for the great work!

Tormak9970:
Awesome, happy to contribute. Hopefully I can help out more in the future.

TrainDoctor (changes requested):
Upon testing on my end, first re-ordering the moved item does not use the correct name. Once that item is placed though, all subsequent re-orderings work as expected.

OuterWinnie:
A couple of things:

- Controls/UI doesn't seems to be very intuitive, i think that you should be able to access the reload/uninstall menu while you are rearranging some plugin. It also doesn't seems very clear what plugin are you moving.

https://streamable.com/txrblf

- If you are reordering a plugin, and moves it really fast, it deselect the one you are moving and changes other different plugin for some reason:

https://streamable.com/s0fm19

AAGaming00:
> A couple of things:
> 
> - Controls/UI doesn't seems to be very intuitive, i think that you shouldn't be able to access the reload/uninstall menu while you are rearranging some plugin. It also doesn't seems very clear what plugin are you moving.
> 
> https://streamable.com/txrblf
> 
> - If you are reordering a plugin, and moves it really fast, it deselect the one you are moving and changes other different plugin for some reason:
> 
> https://streamable.com/s0fm19
> 
> - It's not saving the order between reboots, it changes to the "default" order after a reboot.

Please upload videos directly to the issue instead of using Streamable or other file hosts.

Tormak9970:
Once you do I'll look into it. Is the plugin order being loaded the first time you boot the loader?

OuterWinnie:
> > A couple of things:
> > 
> > * Controls/UI doesn't seems to be very intuitive, i think that you shouldn't be able to access the reload/uninstall menu while you are rearranging some plugin. It also doesn't seems very clear what plugin are you moving.
> > 
> > https://streamable.com/txrblf
> > 
> > * If you are reordering a plugin, and moves it really fast, it deselect the one you are moving and changes other different plugin for some reason:
> > 
> > https://streamable.com/s0fm19
> > 
> > * It's not saving the order between reboots, it changes to the "default" order after a reboot.
> 
> Please upload videos directly to the issue instead of using Streamable or other file hosts.

Sorry, i didn't know i could upload them here directly. I have change it already.

TrainDoctor (approved):
Recent changes has it working well, with a odd issue I'm having other users test to see if they can replicate. Approving for now unless something drastically changes.

TrainDoctor (commented):
Suggestion to hopefully resolve an intermittent issue/hard to reproduce issue. Looks like not all plugins are getting saved in the proper order after a first reboot, and sometimes a few after second reboot as well, but past those it seems to be fine. Possibly an issue with how plugin-order is getting saved, and whether all plugins are getting added to the list until they are interacted with. I don't have anything to back this up as it's just a hunch, but seems to make sense based on manipulation of the settings.json file to replicate what I was seeing.

Tormak9970:
> Suggestion to hopefully resolve an intermittent issue/hard to reproduce issue, checks to see if all plugins installed are in the plugin-order on start/shutdown of decky and at a reasonable interval. Looks like not all plugins are getting saved in the proper order after a first reboot, and sometimes a few after second reboot as well, but past those it seems to be fine. Possibly an issue with how plugin-order is getting saved, and whether all plugins are getting added to the list until they are interacted with. I don't have anything to back this up as it's just a hunch, but seems to make sense based on manipulation of the settings.json file to replicate what I was seeing.

If I understand correctly, there is an issue with the plugin order not saving after the user reorders, saves, and restarts? And this is reflected in the settings.json?

TrainDoctor (approved):
Reordering finally makes it's way to Decky Loader.

PR #353, opened 2023-01-23
Expose a 'decky_plugin' module to decky plugins

popsUlfr:
This pull request adds support for exposing a decky python module to help plugins with useful constants and functions.
There's support for a `_migration` to be exposed in the `Plugin` class to move files to the recommended locations.

*Why not just do it in the `_main` ?* Even though right now it's simply executed right before `_main` by having the migrations in a separate method it could give the loader more flexibility in the future to process them all before the plugins themselves or treat their execution differently.

The `plugin/` folder represents the public interface of the loader to be used by the plugins and is to be included as additional data folder in the resulting loader executable. The path to the public interface is injected just before loading the plugin module as well as the path to `py_modules` relative to the plugins' root for additional dependencies support.

I'm following this up with a pull request to the decky-plugin-template to simply show how to use the decky module. Basically for proper intellisense in the plugin folder all that is needed is to add the path to decky-loader's `plugin/` path to `python.analysis.extraPaths` and you're good to go.

An additional environment variable is exposed that can prove useful for moving data:
- `DECKY_USER_HOME`: the home of the user where decky resides in.
  + e.g.: `/home/deck`

Tell me what you think about this, I think it's a simple enough addition to keep plugin devs' code "tight" and to the point by relying on a common plugin library. And is a starting point make the loader more flexible to logic changes while keeping existing plugins working.

PR #387, opened 2023-02-26
Add cross-platform support to decky

suchmememanyskill:
This PR overhauls decky's backend code to make it cross compatible on windows. Notable changes include:
- Removing all inline cli calls to chown/chmod/systemctl in favor of wrappers managed by the `localplatform.py` file.
    - Chown now does not accept specific user id's anymore, but rather an abstraction in the form of `UserType.HOST_USER`, `UserType.EFFECTIVE_USER` and `UserType.ROOT`.
- Rewrite the logic of sockets for `plugin.py` so that multiple socket types are supported.
    - Note: This also seperates all socket logic into `socket.py`, making `plugin.py` just focused on loading an managing plugins.
- The action workflow has been edited to automatically build a windows binary.
- `updater.py` now downloads the new PluginLoader binary first, then moves it to the right place. This was required for windows to work (you cannot modify a running executable in windows).
- From the https://github.com/SteamDeckHomebrew/decky-loader/pull/353 PR,
    - Imports were moved to main.py, so they don't run multiple times.
    - `get_system_pythonpaths()` has been overhauled to use newlines for splitting paths and to add a failsafe.
    
    
This PR messes with a lot of the core functionality of decky. I have tested everything in this PR, on both windows and steamos and everything *appears* to run fine. Requesting your re-evaluation to double check my work. Please also review the changed code to make sure i didn't do anything weird. I'm sure this PR is nowhere near final yet, but i'd love to hear your feedback.

**For normal users, please do not ask for how to run this.**

Issues to work out still:
- Websockets are abysmally slow on windows. A workaround may be needed for this


TrainDoctor (approved):
Current cursory testing seems to indicate working as intended. I'll test this PR for the rest of the weekend and if I and others don't encounter any issues then this will likely be merged late sunday. Thanks for the great work @suchmememanyskill.

doZennn:
I would like to propose a new field in plugin.json to specify the supported platforms and for the store to filter unsupported plugins out based on that.

TrainDoctor:
> I would like to propose a new field in plugin.json to specify the supported platforms and for the store to filter unsupported plugins out based on that. And possibly for the loader to not attempt to load a plugin unless it's the correct platform as a safety net.
> 
> ~This is probably the wrong place for this but i figured it would be okay here since i couldn't find a main place to discuss multiplat stuff~

This is a good idea but I don't think we have plans to support decky for windows via any time soon. Please make this suggestion again when we start making a minor features suggestion project for plugin api v2.

TrainDoctor:
No objections to this have been raised, merging into decky.

PR #389, opened 2023-03-04
Expose USER_ID and DECKY_USER_ID to plugins

jvyden:
A situation came up in [my plugin](https://github.com/jvyden/deck-kdeconnect/blob/main/main.py#L7-L22) where I needed to get the user's id, and saw no variable in `decky_plugin` for it, so I added it using existing APIs.

Normally this should be `1000` but it can technically be different, so here we are.

*P.S.: This is unfortunately untested due to complications regarding getting Decky running locally, so it would be appreciated if someone could give this a proper test :pray:*

### DIFF
--- a/backend/plugin.py
+++ b/backend/plugin.py
@@ -64,8 +64,10 @@ class PluginWrapper:
             # export a bunch of environment variables to help plugin developers
             environ["HOME"] = helpers.get_home_path("root" if "root" in self.flags else helpers.get_user())
             environ["USER"] = "root" if "root" in self.flags else helpers.get_user()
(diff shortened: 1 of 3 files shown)


TrainDoctor:
If you can make a PR for the plugin template to update the .pyi stub then I can approve and merge both changes. Cheers.

TrainDoctor (approved):

TrainDoctor:
@jvyden apologies, i have had to revert this merge and some changes I made along with it due to issues it introduced (broke plugins both as int or as str). I'd be happy to accept a PR that's been tested (again totally my fault for accepting the PR w/o having tested it).

PR #395, opened 2023-03-09
Shared Ctx tab rename to SharedJSContext

suchmememanyskill:
image
Latest beta update seems to have renamed the shared steam context tab to 'SharedJSContext'. Injecting current frontend code into that tab seems to be working fine

### DIFF
--- a/backend/injector.py
+++ b/backend/injector.py
@@ -394,8 +394,10 @@ async def get_tab_lambda(test) -> Tab:
         raise ValueError(f"Tab not found by lambda")
     return tab
 
+SHARED_CTX_NAMES = ["SharedJSContext", "Steam Shared Context presented by Valve™", "Steam", "SP"]
+
 def tab_is_gamepadui(t: Tab) -> bool:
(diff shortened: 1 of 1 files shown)


backend/injector.py
```diff
@@ -415,4 +415,4 @@ async def close_old_tabs():
         if not t.title or (t.title != "Steam Shared Context presented by Valve™" and t.title != "Steam" and t.title != "SP"):
```
beebls, line 415: `SharedJSContext` should also be added to this line

beebls (changes requested):
Just one more place it needs to be, then I think it should be good for testing and merging (won't be able to test for a few hours)

beebls (approved):

wonderwallwoody:
Works well for me on Beta channel. Decky shows up in QAM and is functional (store, settings, and plugin install works). Here are the logs from the test after a full restart:
[deckylogs.txt](https://github.com/SteamDeckHomebrew/decky-loader/files/10933698/deckylogs.txt)

wonderwallwoody:
Also seems to work fine on Stable channel. Attached are the logs from a full restart and plugin install while on Stable
[deckylogs_stable.txt](https://github.com/SteamDeckHomebrew/decky-loader/files/10934421/deckylogs_stable.txt)

PR #400, opened 2023-03-21
Update dev dependencies to their corresponding latest version

RodoMa92:
This is just a standard upstream for the libraries included for the development of the project. AFAICS this do not cause any issues in the project, but it should be checked from other people just to be sure.


frontend/src/components/modals/filepicker/index.tsx
```diff
@@ -128,6 +128,7 @@ const FilePicker: FunctionComponent<FilePickerProps> = ({
                     ) : (
                       <div style={iconStyles}>
                         {file.realpath.includes('.') ? (
+                          //@ts-ignore
```
AAGaming00, line 131: Why the ignore?

AAGaming00 (commented):
react-router types also should really be pinned to the version Steam uses, original was wrong too tho

PR #416, opened 2023-04-04
Optimize filepicker 

suchmememanyskill:
Current filepicker_ls function could be better by adding pagination, using os.scandir vs os.listdir for slightly faster file checking and adding limits to not lock everything up.

### DIFF
--- a/backend/utilities.py
+++ b/backend/utilities.py
@@ -181,30 +181,30 @@ class Utilities:
         await service_stop(helpers.REMOTE_DEBUGGER_UNIT)
         return True
 
-    async def filepicker_ls(self, path, include_files=True):
+    async def filepicker_ls(self, path, include_files : bool = True, max : int = 1000, page : int = 1):
         # def sorter(file): # Modification time
         #     if os.path.isdir(os.path.join(path, file)) or os.path.isfile(os.path.join(path, file)):
         #         return os.path.getmtime(os.path.join(path, file))
         #     return 0
         # file_names = sorted(os.listdir(path), key=sorter, reverse=True) # TODO provide more sort options
-        file_names = sorted(os.listdir(path)) # Alphabetical
 
-        files = []
+        items = list(os.scandir(path))
+        files = sorted([x for x in items if x.is_file()], key=lambda x: x.name)
+        folders = sorted([x for x in items if x.is_dir()], key=lambda x: x.name)
 
-        for file in file_names:
-            full_path = os.path.join(path, file)
-            is_dir = os.path.isdir(full_path)
-
-            if is_dir or include_files:
-                files.append({
-                    "isdir": is_dir,
-                    "name": file,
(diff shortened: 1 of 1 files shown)


suchmememanyskill:
Pull request is a draft until AA has had the time to work on this for the frontend as well

SkyLeite (changes requested):
Nice, pagination should make this less expensive to run.

One small request though. I usually wouldn't care about this, as I tend to prefer list comprehensions myself, but since the point of this PR is optimizing for when there are too many files, I think cutting down the amount of times we iterate over the entries is worth the readability penalty of using a for loop.

```py
# Pseudo-code, not tested
files, folders, all = [], [], []
for item in items:
    if item.is_file() and include_files:
        files.push(to_dict(item, False))
    elif item.is_dir():
        folders.push(to_dict(item, True))
    else:
        continue

all = sorted(folders + (files if include_files else []))
```

It also lets you skip loading all the entries in memory at once with `items = list(os.scandir(path))`, which can be problematic when there are too many entries, by changing it to `items = os.scandir(path)`

suchmememanyskill:
This wouldn't work the same: folders are placed before files are, something which you cannot do with the suggested pseudocode. I do like the idea of looping only once here.

I do wonder if the os.scandir is the slowest bit of all of this. Os.path.realpath gets called for each entry with no real benefit i think.

PartyWumpus:
Decided to test `list(os.scandir)`'s speed using timeit.timeit:
1,000 files: 0.018s
10,000 files: 0.21s
50,000 files: 1.3s
100,000 files: 2.6s
(nearly linear)

However, timing it with two `time.time()`s makes `list(os.scandir)` run significantly faster, and the sorting functions at the end take up most of the time. The total time adds up to less than what `timeit.timeit` would predict, and the file reading is definitely all happpening when `list(os.scandir)` happens, so i think all my testing was pointless.
But I already did the testing so I'm posting it here anyway!

Total Time:
1,000 files: 0.13s
10,000 files: 0.13
50,000 files: 0.67s
100,000 files: 1.4s

PartyWumpus:
Okay ignore my previous comment.
After further testing, the slowest part of the operation is most certainly os.path.realpath, taking up ~96% of the processing time.


removing realpath speeds it up to 0.7s for 1,000,000 files, so more than fast enough. 

A one line version `sorted([to_dict(x, x.is_dir()) for x in items], key=lambda x: (not(x['isdir']), x['name']))` and skyLeite's code modified appear to be about the same speed as before if not slower.

Yes this was completely unneeded but I wanted to learn how to profile python.

SkyLeite (approved):
I still feel like we'd benefit from reducing the amount of iterations of the files, but if it's faster it's faster. We can always come back to it later if we need to optimize it further :)

Good job!

suchmememanyskill:
Closed in favour of #454

PR #424, opened 2023-04-13
Refactory of the python code

Domefemia:


backend/updater.py
```diff
@@ -132,16 +132,12 @@ async def version_reloader(self):
...
-
-        for x in self.remoteVer["assets"]:
-            if x["name"] == download_filename:
-                download_url = x["browser_download_url"]
-                break
-        
-        if download_url == None:
+        download_temp_filename = f"{download_filename}.new"
+
+        download_url = next((x["browser_download_url"] for x in self.remoteVer["assets"] if x["name"] == download_filename), None)
```
PartyWumpus, line 138: I think this line is less readable than before

TrainDoctor:
While I appreciate the PR, most of this could be accomplished through linting etc. If you would like to submit a PR that adds linting and a code formatting guide that would be of a great benefit. Your submission is lovely but we won't be merging it at this time.

PR #429, opened 2023-04-20
Make patch notes modal only show current branch

beebls:
Spent *way* too long on this, but hey its done and i can finally move onto other things.

Currently, this achieves this by filtering the `remoteVersions` list in python, however it could eventually be refactored to only filter what is returned to the frontend, that way the backend always has the full dataset


TrainDoctor (approved):
Tested and working well, sending to production.

PR #432, opened 2023-04-23
[RFC] Add info/docs page for each plugin

PartyWumpus:
image

image

image

current folder layout is this
```
~/homebrew/plugins/[pluginname]
├── docs
│   ├── config.json
│   ├── en-US
│   │   ├── docs.json
│   │   ├── file1.md
│   │   └── file2.md
│   └── fr-FR
│       ├── docs.json
│       ├── file1.md
│       └── file2.md
├── LICENSE
├── main.py
...
└── README.md
```
docs.json:
```json
{
    "file1.md": "Title of this page",
    "file2.md": "Title of this page"
}
```
config.json: (These are the default values if the file is missing or incomplete)
```json
{
    "default_language": "en-US",
    "include_readme": "False"
}
```

If the docs folder is missing, it'll just display the readme file.
If a page is present in the `default_language`'s folder, but not in the `resolvedLanguage`'s folder, then it'll display the default language's page.

changes that I still need to make are:
+ If there is only one page, don't use `SidebarNavigation`, just display the page as a full screen.
+ Add extra config options like not showing the title.
+ Change the way the cursor moves over the info button so you have to go right from the back button (currently it acts as if the info button is below the back button). Probably involves putting a `Focusable` somewhere.
+ Maybe add a setting somewhere to allow devs to completely disable the info button for their plugin.

This change assumes the i18n changes have already been merged (uses i18next to get the current language in `Docs.tsx`), so should not be merged until then. For testing I suppose I could remove that and replace it with a hardcoded `en-US`.


JohnnyCrazy:
Related to https://github.com/SteamDeckHomebrew/decky-loader/pull/432

PartyWumpus:
> What do you think about moving the idea of `docs.json` into the [frontmatter](https://jekyllrb.com/docs/front-matter/) of each file?

This sounds like a very good idea, I considered something similar but wasn't sure exactly how to implement it. Would also probably make the logic of the loading of files a little simpler.

NGnius:
From the technical side, I can see a couple of downsides to that:
- Every file must be read to build the list of titles for the frontend's side menu (bad scalability)
- Frontmatter is valid markdown (risk of parsing/display bugs)

Both are solvable problems, just things to keep in mind. The advantages seem to outweigh those challenges.

Since it's only the title, is there any reason to not just always use <title>.md? I.e. use the filename as the title and just remove the file extension. Filenames can have spaces and most special characters in Linux so it's not too hard to write titles with good punctuation (unlike Windows).

PartyWumpus:
> Since it's only the title, is there any reason to not just always use <title>.md? 

I want files to be able to fall back to the original language's version of that file if they don't exist in your language. If they're named that way there's no clear way to fall back for individual files. If it isn't considered particularly important, then that could be fine.

Also, I just remembered the reason why I didn't do it the frontmatter way, because it gives no clear order and I'd like devs to be able to order their pages, and if possible add seperators between pages in the list. I think a consistent order is probably reasonably important.

JohnnyCrazy:
> Also, I just remembered the reason why I didn't do it the frontmatter way, because it gives no clear order and I'd like devs to be able to order their pages, and if possible add seperators between pages in the list. I think a consistent order is probably reasonably important.
I suppose I could still use frontmatter for the title and move the ordering done by docs.json into the singular config file as an (optional) list.

Yes, I agree the navbar ordering would be best inside a top-level config file, like `config.json`, as it should be language independent. 

> From the technical side, I can see a couple of downsides to that:
> 
> * Every file must be read to build the list of titles for the frontend's side menu (bad scalability)

I think reading the titles into some struct can be a one-time effort, e.g. when the plugin is loaded. We're also talking about docs pages, which will propably never go beyond 10 per plugin, so I don't think we run into scalability issues here.

ReanimationXP:
Just thought I'd throw in that "perfect is the enemy of good" here. This ultimately is a basic feature and personally I'd settle for a single file readme.md for each plugin. Honestly that should suffice (and should be required, imo), rather than a multipage multifile tabulated novel.  Translations are understandable but beyond that, it's a plugin of a plugin. If you need to write a book about it, just link to your github. My 2 cents you didn't ask for ;)

TrainDoctor:
Wanting to know if there will be any more work on this RFC? The feature is a great idea but I worry that adapting it into the current codebase and getting plugin authors to utilize it would be difficult.

PartyWumpus:
I think this PR was over scoped, it would've been nice if I'd finished it but I think it does too much. An easy way to include arbitrary files (like markdown files) is probably the better way of making the doc writing experience better. The new plugin API might already do this even?

TrainDoctor:
> I think this PR was over scoped, it would've been nice if I'd finished it but I think it does too much. An easy way to include arbitrary files (like markdown files) is probably the better way of making the doc writing experience better. The new plugin API might already do this even?

I do not believe we have arbitrary file support yet but I think most plugins have what they need at the moment. Closing for now with the understanding that this doesn't fit with how most of plugin devs see plugins working going forward.

PR #439, opened 2023-05-03
Hotfix for i18n where the detector was overriding localStorage

RodoMa92:
This fixes the localStorage issues with the plugins.


PartyWumpus:
Appears to fix vibrantdeck, game theme music and protondb badges (which are the only plugins I know were affected). Works on my deck (with stable steamos/client) and on my desktop. This is probably good to merge and post another prerelease.

nabel0:
can confirm the same

PR #449, opened 2023-05-10
Bump DFL to 3.20.7

PartyWumpus:


PartyWumpus:
Everything appears to be unchanged on stable client (good).
Decky is working on beta client, but every single plugin looks to need a DFL update...
image
image

PR #454, opened 2023-05-16
[Feature] File picker improvements

RodoMa92:
This PR implements the additional features from @doZennn into Decky Loader, adding translation on top and adding a few features (like a nicer way to signal errors to the user and additional ways for the requester to specify if they want to allow all file types to be selected or not).


TrainDoctor:
@SteamDeckHomebrew/decky-testers please give this a go.

nabel0 (changes requested):
No mater if i choose to show only ".zip" files or not. The Filepicker always only show directorys and not individual files.

RodoMa92:
I've fixed the issues, missed that in python splitext return also the leading dot for the extension. Now the filtering seems to work fine, hovewer the multimodal seems to not be rerendered after modifying from the filepicker. If @doZennn can give a quick look to understand why it doesn't trigger a rerender, it would be extremely helpful.

In any case, it's just a visual bug and not a functionality bug.

Marco.

frontend/src/components/settings/pages/developer/index.tsx
```diff
@@ -17,21 +17,10 @@ import { useSetting } from '../../../../utils/hooks/useSetting';
 import RemoteDebuggingSettings from '../general/RemoteDebugging';
 
 const installFromZip = () => {
-  window.DeckyPluginLoader.openFilePicker('/home/deck', true).then((val) => {
+  window.DeckyPluginLoader.openFilePicker('/home/deck', true, true, ['zip'], false, false).then((val) => {
```
TrainDoctor, line 20: Let's change this from a hardcoded directory to one relative to the current user if possible.

frontend/src/components/settings/pages/developer/index.tsx
TrainDoctor, line 1: File needs conflicts resolved so that it can be tested.

TrainDoctor (changes requested):
See comments, code appears to be largely fine but I think it's important we don't use a hardcoded path. Other than that great job and once fixes are made/conflicts are resolved and this can be tested it should be good to merge.

frontend/src/components/settings/pages/developer/index.tsx
RodoMa92: Besides using dumb things like opening ., I would resolve the environment variables for respecting the home directory, however I can't find any way to do this frontend side. It's correct that it's not exposed by default on the frontend by default and is only accessible on the backend?

RodoMa92: It's fixed on the latest commit, just need to be tested.

nabel0 (approved):
Works fine for me

PR #456, opened 2023-05-17
Added french localization.

elliotfontaine:


elliotfontaine:
Just saw you're using the Weblate API now. Should I rather start a new translation there ?

TrainDoctor:
> Just saw you're using the Weblate API now. Should I rather start a new translation there ?

Yes that would be ideal.

elliotfontaine:
> > Just saw you're using the Weblate API now. Should I rather start a new translation there ?
> 
> Yes that would be ideal.

Done.

PR #462, opened 2023-05-20
Add env var to not replace systemd service file

suchmememanyskill:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [X] This is a new feature

# Description

This adds an environment variable that stops decky from updating it's systemd unit/config, so any configuration in it will not be lost. Default behaviour stays the same: the systemd unit/config gets updated every time decky updates

Tl;dr: Simply add `Environment=KEEP_SYSTEMD_SERVICE=1` to decky's systemd unit/config to stop it from being overwritten

PS: This also fixes a bug where the LIVE_RELOAD environment variable was ineffective

### DIFF
--- a/backend/localplatform.py
+++ b/backend/localplatform.py
@@ -32,7 +32,10 @@ def get_server_port() -> int:
     return int(os.getenv("SERVER_PORT", "1337"))
 
 def get_live_reload() -> bool:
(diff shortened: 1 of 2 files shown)


TrainDoctor (approved):
Tested, working. Merging.

PR #499, opened 2023-06-29
Make sure settings/data/logs gets created as user

suchmememanyskill:
Please tick as appropriate:
- [X] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

This fixes the imfamous 'emudeck' plugin loading issue (which turns out, is not caused by emudeck). When installing a root plugin as the first plugin, any non-root plugins installed after will fail to load. This PR fixes that


PartyWumpus:
works for me

PR #502, opened 2023-07-08
Add descriptions to Decky titleview DialogButtons

FeroTheFox:
- [ ] This is a bugfix/hotfix
- [X] This is a new feature

# Description
This simple PR is in preparation for custom titleviews. Since plugins may reuse the same icons or use ones that aren't universally recognized, it will be a good practice to disambiguate Decky's buttons with labels. This also has precedence in the Steam UI, where any icon-only button has a matching description.

### DIFF
--- a/backend/locales/en-US.json
+++ b/backend/locales/en-US.json
@@ -234,6 +234,10 @@
       "testing": "Testing"
     }
   },
+  "TitleView":{
+    "decky_store_desc": "Open Decky Store",
+    "settings_desc": "Open Decky Settings"
+  },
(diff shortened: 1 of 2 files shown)


PartyWumpus:
This should probably use the translation stuff instead of being hardcoded, I'll see if I can do that tomorrow

RodoMa92:
> This should probably use the translation stuff instead of being hardcoded, I'll see if I can do that tomorrow

Already done :P

PartyWumpus:
lgtm, thanks

Issue #446, opened 2023-05-08
[BUG] decky-loader doesn't install plugins


metya:
### Please confirm

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] I have checked the [common issues section in the readme file](https://github.com/SteamDeckHomebrew/decky-loader#-common-issues)
- [X] I have attached logs to this bug report (failure to include logs will mean your issue will not be responded too).

### Bug Report Description

I have a very strange bug.

I installed decky-loader within `decky-installer.desktop` and it seems to work, but the store isn't working properly - there are no pictures and the install button doesn't do anything.  I noted that the website https://plugins.deckbrew.xyz/ doesn't work, but I can retrieve json from https://plugins.deckbrew.xyz/plugins, and even can retrieve files from cdn address like `https://cdn.tzatzikiweeb.moe/file/steam-deck-homebrew/versions/968444702eaf59e1c107922a01d5028f07d4e4753a6f4f6aa2948b00e02a044f.zip` for example. 

When I try to install some plugin in the logs there are line about it. 

```
May 08 16:36:21 steamdeck PluginLoader[54108]: [browser][INFO]: Installing Decky Cloud Save from URL (Version: 1.2.1)
May 08 16:36:21 steamdeck PluginLoader[54108]: [browser][DEBUG]: Fetching https://cdn.tzatzikiweeb.moe/file/steam-deck-homebrew/versions/968444702eaf59e1c107922a01d5028f07d4e4753a6f4f6aa2948b00e02a044f.zip
```

But nothing changes, and it seems that downloading doesn't happen, cause `aiohttp_session` is still open when you start to do something else, and it closes after that with a log like 
```
May 08 16:34:35 steamdeck PluginLoader[47597]: [base_events][ERROR]: Unclosed client session
May 08 16:34:35 steamdeck PluginLoader[47597]: client_session: <aiohttp.client.ClientSession object at 0x7fe73522f610>
``` 

So nothing installs to the plugin's directory. 

But if download the zip file manually and unzip it in that directory, that works as it should. A plugin is in the list and works properly. 

### Expected Behaviour

It should work just fine :)

### SteamOS version

Steam OS 3.4.6 Stable

### Selected Update Channel

Stable

### Have you modified the read-only filesystem at any point?

No

### Logs

[deckylog.txt](https://github.com/SteamDeckHomebrew/decky-loader/files/11423446/deckylog.txt)

[...]

PR #510, opened 2023-07-21
Ignore chmod if decky is not run as root

PartyWumpus:
Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

This fixes the issue issue mentioned here https://github.com/Jovian-Experiments/Jovian-NixOS/pull/111 and here https://github.com/SteamDeckHomebrew/decky-loader/issues/446#issuecomment-1637177041

While decky doesn't really support running as the user, as it breaks some of its features, we might as well make it functional where possible. This *has not been tested*, i'll test it in a moment and then merge it myself, so I don't really know who I'm writing this for or why I'm making it a PR instead of just commiting to main...

### DIFF
--- a/backend/localplatformlinux.py
+++ b/backend/localplatformlinux.py
@@ -1,4 +1,4 @@
-import os, pwd, grp, sys, logging
+import os, pwd, grp, sys, logging, geteuid
 from subprocess import call, run, DEVNULL, PIPE, STDOUT
 from customtypes import UserType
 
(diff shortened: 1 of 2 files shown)


backend/localplatformlinux.py
```diff
@@ -60,6 +60,8 @@ def chown(path : str,  user : UserType = UserType.HOST_USER, recursive : bool =
     return result == 0
 
 def chmod(path : str, permissions : int, recursive : bool = True) -> bool:
+    if os.geteuid != 0:
```
samueldr, line 63: You probably also want to log a warning for future support issues, allowing the project support peeps to understand that it's not running as root, I'd assume.

PartyWumpus: Probably best to just show this once on startup, but yes that is a good idea, will do

samueldr: Sure, I'm about 0% familiar with the codebase, the overall important thing is to ensure you all in support position can identify the situation, now that it can happen.

PR #512, opened 2023-07-23
Add Custom `TitleView`

beebls:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This is an addition that allows plugin developers to change the titleView of their plugin. Currently, it is just a `<div style={{ flex: 0.9 }}>{activePlugin.name}</div>`. If a plugin has the `titleView` prop in their definePlugin return value, it will now display that instead of the boilerplate name.

This does not allow for removal of the back button, as that is crucial to ensuring people can exit out of plugins, the custom titleView slots in to the right of the back button.


PartyWumpus:
image
Tested and it works fully.

PR #515, opened 2023-07-27
[Need Testing] Actually fix sqlite 3 issues

RodoMa92:
Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

With this build chain change now the bundled SQLite binary should be the latest and greatest and the issue with SDH-PlayTime should be put to rest properly.


RodoMa92:
This actually fixed any issues with PlayTime, just tested now.

PR #525, opened 2023-08-11
Only keep up to 5 recent logs of runs of plugins

suchmememanyskill:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [X] This is a new feature

# Description

Currently, decky plugins utilising decky_plugin.py endlessly write logs to the same file, eventually leading to gigantic log files. This fixes that by naming the log file after the current time. Effectively, every time a plugin is ran a new log file is created. This PR also deletes any .log file older than the 5 newest files. Example new log name: `2023-08-11 14.50.16.log`

### DIFF
--- a/plugin/decky_plugin.py
+++ b/plugin/decky_plugin.py
@@ -17,6 +17,7 @@ __version__ = '0.1.0'
 import os
 import subprocess
 import logging
+import time
 
 """
 Constants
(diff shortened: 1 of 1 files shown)


PartyWumpus:
LGTM and works well on deck & my pc, I assume you tested it on windows, because I'm not going to.

Issue #513, opened 2023-07-26
[BUG] Plugin order is not consistent after updating plugins


PartyWumpus:
### Please confirm

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] I have checked the [common issues section in the readme file](https://github.com/SteamDeckHomebrew/decky-loader#-common-issues)
- [ ] I have attached logs to this bug report (failure to include logs will mean your issue will not be responded too).

### Bug Report Description

After a plugin is updated, the plugin order is succesfully restored to the order it should be, but then after the steam client is restarted or if decky is restarted, the updated plugins move to the bottom. If the order is manually changed before the client restarts, it'll work fine.

The frontend has the correct order, but the backend does not. When a plugin is updated the loader.json's pluginOrder list first loses that plugin from the list and then it appears again at the bottom. This is likely fixable either by putting the same code that fixes the order in the frontend into the backend as well, or just saving the order from the frontend whenever a plugin is updated or possibly removing the code that's incorrectly saving to the loader.json file on a plugin update.

### Expected Behaviour

The order should be consistent. 

### SteamOS version

SteamOS 3.5 Main

### Selected Update Channel

Stable

### Have you modified the read-only filesystem at any point?

Yes, not relevant

### Logs

not relevant

btabaska:
#522 Pull Request to add backend settings sync to plugin order like notification settings uses.

nabel0:
I can confirm this...Same here

PR #530, opened 2023-08-25
Preserve plugin order when reinstalling/updating

jurassicplayer:
Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

This fixes issue: #513

This change preserves the plugin order if the plugin is already installed and assumes that when calling _install on an already installed plugin, either reinstall/update was used and the entry should in theory still be there after install (this assumption can fail if the install fails). This should preserve the order across __full reboots__ and should be tested with full reboots after plugin changes (re-installs and updates, but no harm in checking plugin installs and deletions as well).

### DIFF
--- a/backend/browser.py
+++ b/backend/browser.py
@@ -139,6 +139,8 @@ class PluginBrowser:
 
         # Check if plugin is installed
         isInstalled = False
+        # Preserve plugin order before removing plugin (uninstall alters the order and removes the plugin from the list)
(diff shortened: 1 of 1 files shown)


nabel0 (approved):
Tested and works fine under SteamOS Stable/Beta/Preview

TrainDoctor (approved):

PR #539, opened 2023-09-13
Call plugin unload function after stopping event loop

Jan200101:
This can prevent race conditions where unload is clearing data but main is still working with it

Please tick as appropriate:
- [X] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

During the unload long-running code is still active causing possible race conditions when unloads cleans up data which main was referencing.

## Example Plugin
```py
class Plugin:
    data = []

    async def _main(self):
        while True:
            self.data.append(len(self.data))
            decky_plugin.logger.info(self.data)
            await asyncio.sleep(0.1)

    async def _unload(self):
        self.data = None

```

### DIFF
--- a/backend/src/plugin.py
+++ b/backend/src/plugin.py
@@ -118,11 +118,11 @@ class PluginWrapper:
 
         if "stop" in data:
             self.log.info("Calling Loader unload function.")
-            await self._unload()
             get_event_loop().stop()
             while get_event_loop().is_running():
                 await sleep(0)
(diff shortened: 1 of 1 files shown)


TrainDoctor:

TrainDoctor:
Going to give this a go. Apologies it took so long to get to this.

TrainDoctor (approved):
Tested out on a local version and worked perfectly. Merged!

PR #542, opened 2023-09-14
Turn command id into global

Jan200101:
Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

the response id is tied to the request id thus multiple instances of the Tab class can cause problems when prior messages are not consumed properly

### DIFF
--- a/backend/src/injector.py
+++ b/backend/src/injector.py
@@ -13,6 +13,9 @@ BASE_ADDRESS = "http://localhost:8080"
 
 logger = getLogger("Injector")
 
+# Request IDs have to be globally unique to prevent accidentally reading the wrong response
+cmd_id = 0
+
 class _TabResponse(TypedDict):
     title: str
     id: str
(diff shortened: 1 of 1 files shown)


AAGaming00:
or do they just echo whatever ID we use, that would be fine then

Jan200101:
the format is similar to jsonrpc, the ID I send is the same I receive
This is used to unique identify request and response pairs.

When dealing with multiple different Tab instances this isn't an issue, but when there are multiple Tabs pointing towards the same websocket the same ID is likely to appear since we don't always consume the stream.

Jan200101:
ignore what I said, the format *is* JSON-RPC 1.0

And with that we should set the ID to None when no response is required to inform the server we don't expect a response

Jan200101:
scratch that, they don't support notifications, only requests.

TrainDoctor:

PR #543, opened 2023-09-14
Use python executable directly instead of guessing

Jan200101:
In the very unlikely case that no executable was found this will fall back onto "python"

Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

Instead of blindly guessing the executable just use the currently interpreter executable with a fallback to plain "python"

### DIFF
--- a/backend/helpers.py
+++ b/backend/helpers.py
@@ -65,7 +65,7 @@ def get_system_pythonpaths() -> list[str]:
         extra_args["env"] = {}
 
     try:
-        proc = subprocess.run(["python3" if localplatform.ON_LINUX else "python", "-c", "import sys; print('\\n'.join(x for x in sys.path if x))"],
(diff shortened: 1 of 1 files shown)


AAGaming00:
i think this would break with pyinstaller builds?

Jan200101:
Yep, it would.
Looking further there appears to be no way around this.
Simply looking for "python" is still a bad way of doing this though.

HighSpeed1380:
I want to know the reason why built using python

mantikafasi:
"High Speed" would rather have electron installer

marios8543:
🤮

TrainDoctor:
> "High Speed" would rather have electron installer

We will never use electron for a goddamn installer... No. Just no.

PR #545, opened 2023-09-15
import decky_plugin during plugin loading and set it up as module

Jan200101:
Please tick as appropriate:
  - This was not tested against the PyInstaller output
- [ ] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

Currently the plugin logic relies on decky_plugin.py being on the path, this PR makes it so that decky_plugin is loaded directly and put into the sys modules ready to be used.

Tested against 2 copies of the plugin template with prints added to verify that the imported instance desky_plugin is indeed different.

Had to import from an absolute path because relative imports can only be done in a package

### DIFF
--- a/backend/src/plugin.py
+++ b/backend/src/plugin.py
@@ -84,6 +84,13 @@ class PluginWrapper:
             for key in keys:
                 sysmodules[key] = sysmodules["src"].__dict__[key]
 
+            # add decky_plugin as module ahead of time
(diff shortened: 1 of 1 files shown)


TrainDoctor (changes requested):
Please update to latest code and this should be an easy merge.
Thanks.

Jan200101:
completely untested both with and without PyInstaller.
Plugin path may be wrong, idk.

Issue #419, opened 2023-04-07
[Request] Decky Plugin Store: add more sort options (show new releases first)


dJani97:
### Please confirm

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] This is not a request for a plugin

### Feature Request Description

As the Decky Plugin Store expands, I think having more sorting options would be nice.

For example:
- newly released
- recently updated
- most popular

As someone who already browsed through the plugin store a couple of times, I'm mostly interested in what's new.

PartyWumpus:
This was resolved by #547, you can now sort by release date and download count. Sorting by most recently updated would also be possible, although it did not seem useful so was not added.

Issue #360, opened 2023-02-02
Decky store does not know what plugins you have installed/out of date


beebls:
### Please confirm

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] I have checked the [common issues section in the readme file](https://github.com/SteamDeckHomebrew/decky-loader#-common-issues)

### Bug Report Description

When browsing through the Decky store, all plugins show an "Install" button, even if they are already installed and up to date, this allows you to "install" already installed plugins.

### Expected Behaviour

There should be some indication of "Installed" or "Out Of Date", either on the install button itself, or somewhere else.
![image](https://user-images.githubusercontent.com/102569435/216202904-8d836b6f-9a04-44c7-ae12-1c71c623540b.png)
This is how CSS/AudioLoader handles it, but any indication would be appreciated.

### SteamOS version

SteamOS 3.4.5 Stable

### Selected Update Channel

Prerelease

### Have you modified the read-only filesystem at any point?

No

### Logs

Not related to backend/logging, purely UI focused.

beebls:
This could be marked either as bug or feature request, I honestly didn't know which it fit more so I just picked bug.

TrainDoctor:
This would generally fall under a feature request. Up until now the store and the rest of the UI have fallen under different contexts and have been treated as such. We might consider implementing something like this, but not at this moment.

beebls:
Understandable, renamed the PR to mirror this.

If I have some time eventually I might try to build something like this and PR it, but currently quite bogged down.

dJani97:
Along with this, an option to only show plugins that are not installed would be nice. I think users go to the store to find new plugins, and such a filter would simplify that process.

TrainDoctor:
> this, it's basic functionality that i expect out of a plugin store and lacking it just feels really janky

This is purely an interface issue. Loader itself is very aware of your plugin updates and this is accessible in the settings menu.

spudpiggy:
> > this, it's basic functionality that i expect out of a plugin store and lacking it just feels really janky
> 
> This is purely an interface issue. Loader itself is very aware of your plugin updates and this is accessible in the settings menu.

yeah i know but it should not be a problem at all

PR #547, opened 2023-10-15
Add sorting by date to the plugin store

PartyWumpus:
Please tick as appropriate:
- [ ] I have tested this code on a steam deck or on a PC -- sort of
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This fixes issue: #419

(This is currently waiting on https://github.com/SteamDeckHomebrew/decky-plugin-store/pull/51 so it actually works instead of just breaking sorting)
Adds two new options to the plugin store's sort box, date ascending and date descending.

### DIFF
--- a/frontend/src/components/store/Store.tsx
+++ b/frontend/src/components/store/Store.tsx
@@ -8,7 +8,7 @@ import {
   TextField,
   findModule,
 } from 'decky-frontend-lib';
-import { FC, useEffect, useMemo, useState } from 'react';
+import { FC, useEffect, useMemo, useState, Dispatch, SetStateAction } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import logo from '../../../assets/plugin_store.png';
@@ -61,7 +61,7 @@ const StorePage: FC<{}> = () => {
   );
 };
 
-const BrowseTab: FC<{ children: { setPluginCount: setPluginCount } }> = (data) => {
+const BrowseTab: FC<{ children: { setPluginCount: Dispatch<SetStateAction<Number>> } }> = (data) => {
 
   const { t } = useTranslation();
 
@@ -79,29 +79,30 @@ const BrowseTab: FC<{ children: { setPluginCount: setPluginCount } }> = (data) =
   const [selectedSort, setSort] = useState<number>(dropdownSortOptions[0].data);
   // const [selectedFilter, setFilter] = useState<number>(filterOptions[0].data);
   const [searchFieldValue, setSearchValue] = useState<string>('');
-  const [data, setData] = useState<StorePlugin[] | null>(null);
+  const [pluginList, setPluginList] = useState<StorePlugin[] | null>(null);
   const [isTesting, setIsTesting] = useState<boolean>(false);
 
   useEffect(() => {
     (async () => {
-      sort, direction = null, null
+      let sort = null
+      let direction = null
       switch (selectedSort) {
-        case 1: direction=SortDirections.ascending;sort=SortOptions.name
+        case 1: direction=SortDirections.ascending; sort=SortOptions.name
         case 2: direction=SortDirections.descending;sort=SortOptions.name
-        case 3: direction=SortDirections.ascending;sort=SortOptions.date
+        case 3: direction=SortDirections.ascending; sort=SortOptions.date
         case 4: direction=SortDirections.descending;sort=SortOptions.date
       }
       const res = await getPluginList(sort, direction);
       logger.log('got data!', res);
-      setData(res);
-      setPluginCount(res.length)
+      setPluginList(res);
+      data.children.setPluginCount(res.length)
       const storeRes = await getStore();
       logger.log(`store is ${storeRes}, isTesting is ${storeRes === Store.Testing}`);
       setIsTesting(storeRes === Store.Testing);
     })();
   }, []);
 
-  return !data ? (
+  return !pluginList ? (
     <div style={{ height: '100%' }}>
       <SteamSpinner />
     </div>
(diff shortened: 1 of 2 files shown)


ReanimationXP:
Please merge this, I really like to keep on top of what the newest plugins are or what has been updated.  It would also be nice if the store would indicate if things are already installed and change the button text to "Update" rather than "Install".  Would make browsing the store a lot less confusing.

PartyWumpus:
> Please merge this.

To be honest, I totally forgot about this because I thought I already had... I'll finish it up and it should be good to merge soon.

> It would also be nice if the store would indicate if things are already installed and change the button text to "Update" rather than "Install". Would make browsing the store a lot less confusing.

Yeah. I'm sure someone will get to it someday https://github.com/SteamDeckHomebrew/decky-loader/issues/360

PartyWumpus:
Alright should be done now.

ReanimationXP:
Thank you!

frontend/src/components/store/Store.tsx
```diff
@@ -49,52 +37,71 @@ const StorePage: FC<{}> = () => {
...
+            },
+          ]}
+        />
       </div>
     </>
   );
 };
 
-const BrowseTab: FC<{ children: { data: StorePlugin[]; isTesting: boolean } }> = (data) => {
+const BrowseTab: FC<{ children: { setPluginCount: Dispatch<SetStateAction<number | null>> } }> = (data) => {
```
JohnnyCrazy, line 64: I know you didn't change this, but it's kinda non-standard to use `children` for such props. We can just pull this up so `{ setPluginCount: ... }` are the props.

frontend/src/components/store/Store.tsx
```diff
@@ -49,52 +37,71 @@ const StorePage: FC<{}> = () => {
...
+      // this is because it was initially done incorrectly for i18n and 'fixing' it would
+      // make all the translations incorrect
+      { data: [SortOptions.name, SortDirections.ascending], label: t('Store.store_tabs.alph_desc') },
+      { data: [SortOptions.name, SortDirections.descending], label: t('Store.store_tabs.alph_asce') },
+      { data: [SortOptions.date, SortDirections.ascending], label: t('Store.store_tabs.date_asce') },
+      { data: [SortOptions.date, SortDirections.descending], label: t('Store.store_tabs.date_desc') },
+      { data: [SortOptions.downloads, SortDirections.descending], label: t('Store.store_tabs.downloads_desc') },
+      { data: [SortOptions.downloads, SortDirections.ascending], label: t('Store.store_tabs.downloads_asce') },
     ],
     [],
```
JohnnyCrazy, line 79: `t` should be a dependency here, but i guess we don't allow language changes yet, so not really required

frontend/src/components/store/Store.tsx
```diff
@@ -49,52 +37,71 @@ const StorePage: FC<{}> = () => {
...
+    })();
+  }, [selectedSort]);
+
+  useEffect(() => {
+    (async () => {
+      const storeRes = await getStore();
+      logger.log(`store is ${storeRes}, isTesting is ${storeRes === Store.Testing}`);
+      setIsTesting(storeRes === Store.Testing);
+    })();
+  }, []);
```
JohnnyCrazy, line 104: Could be a `useMemo`, as `setIsTesting` is not called somewhere else.

frontend/src/components/store/Store.tsx
```diff
@@ -49,52 +37,71 @@ const StorePage: FC<{}> = () => {
...
+  const [isTesting, setIsTesting] = useState<boolean>(false);
+
+  useEffect(() => {
+    (async () => {
+      const res = await getPluginList(selectedSort[0], selectedSort[1]);
+      logger.log('got data!', res);
+      setPluginList(res);
+      data.children.setPluginCount(res.length);
+    })();
+  }, [selectedSort]);
```
JohnnyCrazy, line 96: Could be a `useMemo`, as `setPluginList` is not called somewhere else.

JohnnyCrazy (approved):
Nice!

frontend/src/components/store/Store.tsx
PartyWumpus: oops... To be honest, I didn't notice I was using the children instead of the props. I'm surprised that works at all, very weird.

frontend/src/components/store/Store.tsx
PartyWumpus: we do allow changing language, but it (iirc) uses valve's language settings.

I don't think it's possible to change language without leaving this menu, so I don't think it's worth making it a dependency (I assume you mean putting it in the list on line 79).

frontend/src/components/store/Store.tsx
PartyWumpus: Using useMemo for web requests seems to be not recommended, because it's run during the render, not after. Plus we want the 'loading' state so it can show a spinner.

JohnnyCrazy: Ohhh true, I totally forgot that usememo is also synchronous only. LGTM!

frontend/src/components/store/Store.tsx
PartyWumpus: Not yet, it's an easy fix but I need to get home and make sure it works first. Should be done about an hour from now.

PartyWumpus:
I've just noticed that the store is erroring when sorting by name (but not the other sorts). This shouldn't be merged until that is sorted.

PartyWumpus:
Okay, fixed that. This has been reviewed so i'm going to merge it.

Issue #527, opened 2023-08-18
[BUG] Plugins installed with zip or link do not show up until decky restarts


PartyWumpus:
### Please confirm

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] I have checked the [common issues section in the readme file](https://github.com/SteamDeckHomebrew/decky-loader#-common-issues)
- [X] I have attached logs to this bug report (failure to include logs will mean your issue may not be responded to).

### Bug Report Description

Install a plugin with zip or with link in the developer settings, it will not show up in the plugins list until decky restarts (restarting the frontend alone does *not* fix it, so it's a problem in the backend). Looking through the code, I can't really tell why this happens, as the function calls look the same. (maybe the InstallType is wrong?)

### Expected Behaviour

The plugin should appear right away.

### SteamOS version

SteamOS 3.5 Main, and my EndevourOS desktop

### Selected Update Channel

Prerelease

### Logs

[deckylog.txt](https://github.com/SteamDeckHomebrew/decky-loader/files/12380494/deckylog.txt)

eXhumer:
Hello. I came across this as well when I had to install TabMaster's new update and did a bit more digging through the code to figure out the problem, it seems to be due to the code in `src/backend/browser.py` backend code, more specifically the `_install` method inside it.

First issue is when the function tries to determine if the plugin is already installed, it fails this check until the ZIP file is named exactly the same as the plugin name

https://github.com/SteamDeckHomebrew/decky-loader/blob/9295e4b038a32cbfa49ac7fffa69a3768f786491/backend/src/browser.py#L166

Any ZIP which is not named exactly as the plugin name will fail this check always and will assume the plugin is not installed already.

https://github.com/SteamDeckHomebrew/decky-loader/blob/9295e4b038a32cbfa49ac7fffa69a3768f786491/backend/src/browser.py#L111-L120

This piece of code is never executed as a result of the failed check

https://github.com/SteamDeckHomebrew/decky-loader/blob/9295e4b038a32cbfa49ac7fffa69a3768f786491/backend/src/browser.py#L206-L212

And this is the final bit of mistake. ZIP is extracted, but since the `find_plugin_folder` fails, it raises AssertionError because of an empty plugin_folder and halts the rest of the code execution.

https://github.com/SteamDeckHomebrew/decky-loader/blob/9295e4b038a32cbfa49ac7fffa69a3768f786491/backend/src/browser.py#L214-L219

[...]

PR #578, opened 2024-01-24
Get plugin name for development ZIP during installation

eXhumer:
Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

Currently the development plugin installation from ZIP / URL is broken partially, due to relying on ZIP name as the plugin name. This commit fixes the issue, by parsing the plugin name from the folder containing `plugin.json` from the ZIP. I tested this on my Steam Deck OLED and can confirm it to fix the issue. I installed a ZIP which previously failed installation and it now works without issue with this commit.

## Related comments on issue

https://github.com/SteamDeckHomebrew/decky-loader/issues/527#issuecomment-1908169324

https://github.com/SteamDeckHomebrew/decky-loader/issues/527#issuecomment-1908947062

This fixes issue: #527

### DIFF
--- a/backend/src/browser.py
+++ b/backend/src/browser.py
@@ -10,6 +10,7 @@ from hashlib import sha256
 from io import BytesIO
 from logging import getLogger
 from os import R_OK, W_OK, path, listdir, access, mkdir
+from re import sub
 from shutil import rmtree
 from time import time
 from zipfile import ZipFile
@@ -162,12 +163,6 @@ class PluginBrowser:
         current_plugin_order = self.settings.getSetting("pluginOrder")[:]
         if self.loader.watcher:
             self.loader.watcher.disabled = True
-        try:
-            pluginFolderPath = self.find_plugin_folder(name)
-            if pluginFolderPath:
-                isInstalled = True
-        except:
-            logger.error(f"Failed to determine if {name} is already installed, continuing anyway.")
 
         # Check if the file is a local file or a URL
         if artifact.startswith("file://"):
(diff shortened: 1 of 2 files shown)


backend/src/browser.py
```diff
@@ -198,6 +192,28 @@ async def _install(self, artifact: str, name: str, version: str, hash: str):
                 if res.status != 200:
                     logger.error(f"Server did not accept install count increment request. code: {res.status}")
 
+        if res_zip and version == "dev":
+            with ZipFile(res_zip) as plugin_zip:
+                plugin_json_list = [file for file in plugin_zip.namelist() if file.endswith("/plugin.json")]
```
FineWolf, line 197: the `plugin.json` should always be one sub folder deep within the zip. The other ones have no incidence in how a plugin is loaded.

```python
plugin_json_list = [file for file in plugin_zip.namelist() if file.endswith("/plugin.json") and file.count("/") == 1]
```

backend/src/browser.py
```diff
@@ -198,6 +192,28 @@ async def _install(self, artifact: str, name: str, version: str, hash: str):
...
+                if len(plugin_json_list) == 0:
+                    logger.fatal("No plugin.json found in plugin ZIP")
+                    return
+
+                elif len(plugin_json_list) > 1:
+                    logger.fatal("Multiple plugin.json found in plugin ZIP")
+                    return
+
+                else:
+                    name = json.loads(plugin_zip.read(plugin_json_list[0]))["name"]
```
FineWolf, line 208: The plugin folder can differ from the plugin name in `plugin.json`. For example, `CSS Loader` installs in the `SDH-CssLoader` folder.

The plugin folder name instead should be derived from the folder name in which `plugin.json` is located within the zip.

```python
name = re.sub(r"/.+$", "", plugin_json_list[0])
```

FineWolf (commented):
Added some comments about how plugins are structured, and how the `plugin.json` `name` property is not reflective of the actual folder name in which the plugin is installed.

eXhumer:
> Added some comments about how plugins are structured, and how the `plugin.json` `name` property is not reflective of the actual folder name in which the plugin is installed.

Thanks for the comments regarding the plugin structure, I have amended changes based on the comments left.

FineWolf:
Sigh, wrong account, but yes. Looks good.

TrainDoctor:
@SteamDeckHomebrew/decky-testers need a tester for this one. Please install this PR's binary and then download a zip for a plugin and then rename that zip however you'd like (just don't give it the same name :P) and check that this works.
If this works, this PR can be merged.

TrainDoctor (approved):
Works like a charm, and also helped me to discover that when you download a plugin zip file from github (built during an actions run etc) it downloads as a redundant zip file... Guess we should add some logic to account for that 🤦🏻

PR #585, opened 2024-02-10
fix(filepicker_ls): use case insensitive matching for file exts

doZennn:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description
Adds case insensitive matching to fix files not showing when using `validFileExtensions` in `openFilePickerV2()` .
Example:
```tsx
<FilePicker
    validFileExtensions={['.png']} // this would match ".png" but not ".PNG"
/>
```


RodoMa92:
Nice catch @doZennn, looks good to me.

TrainDoctor (approved):

PR #610, opened 2024-03-28
Add new user agent

suchmememanyskill:
Please tick as appropriate:
- [X] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

This fixes issue: The new beta steam client changes the user agent. This causes decky to close browser tabs it shouldn't. This PR fixes that

### DIFF
--- a/backend/src/injector.py
+++ b/backend/src/injector.py
@@ -412,7 +412,7 @@ async def get_tab_lambda(test: Callable[[Tab], bool]) -> Tab:
 
 SHARED_CTX_NAMES = ["SharedJSContext", "Steam Shared Context presented by Valve™", "Steam", "SP"]
 CLOSEABLE_URLS = ["about:blank", "data:text/html,%3Cbody%3E%3C%2Fbody%3E"] # Closing anything other than these *really* likes to crash Steam
-DO_NOT_CLOSE_URLS = ["Valve Steam Gamepad/default", "Valve%20Steam%20Gamepad/default"] # Steam Big Picture Mode tab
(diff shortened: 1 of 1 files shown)


backend/src/injector.py
```diff
@@ -412,7 +412,7 @@ async def get_tab_lambda(test: Callable[[Tab], bool]) -> Tab:
 
 SHARED_CTX_NAMES = ["SharedJSContext", "Steam Shared Context presented by Valve™", "Steam", "SP"]
 CLOSEABLE_URLS = ["about:blank", "data:text/html,%3Cbody%3E%3C%2Fbody%3E"] # Closing anything other than these *really* likes to crash Steam
-DO_NOT_CLOSE_URLS = ["Valve Steam Gamepad/default", "Valve%20Steam%20Gamepad/default"] # Steam Big Picture Mode tab
+DO_NOT_CLOSE_URLS = ["Valve Steam Gamepad/default", "Valve%20Steam%20Gamepad"] # Steam Big Picture Mode tab
```
EMERALD0874, line 415: Seems like we accept multiple of these. Any reason you chose to remove the previous one rather than add a new option? Just a little concerned about backward compatibility.

AAGaming00: We've asked the question in the discord, it's because `Valve%20Steam%20Gamepad/default` includes `Valve%20Steam%20Gamepad` so it will trigger anyway.

EMERALD0874 (approved):
LGTM given @AAGaming00's comment

PR #620, opened 2024-05-05
[needs websockets] Add development flake file

PartyWumpus:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

Adds a flake.nix file so developers can get all relevant dependencies by just typing `nix develop`. 
Uses poetry2nix so all python deps will get installed at the correct versions without needing to update the flake. 
It's currently set up to build each python library from source instead of using the wheel but this can be changed if preferable.
Possibly an issue is that the env is not automatically updated when new deps are added, it must be restarted.
Also the way pyright is working is a bad solution.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,44 @@
+{
+  description = "Decky development environment";
+  # pulls in the python deps from poetry
+
+  inputs = {
+    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    poetry2nix = {
+      url = "github:nix-community/poetry2nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = { self, nixpkgs, flake-utils, poetry2nix }:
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -0,0 +1,44 @@
...
+    poetry2nix = {
+      url = "github:nix-community/poetry2nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = { self, nixpkgs, flake-utils, poetry2nix }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = nixpkgs.legacyPackages.${system};
```
AAGaming00, line 17: seemingly unnecessary usage of legacyPackages? should be able to call nixpkgs with the system argument instead.

AAGaming00 (changes requested):

flake.nix
PartyWumpus: good point, fixed.

AAGaming00 (approved):
LGTM, could probably add a .envrc too for direnv

Issue #629, opened 2024-05-27
[Request] Ablility to rename plugins


oshata:
### Please confirm

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] This is not a request for a plugin

### Feature Request Description

Have the ability to rename a plugin so that when you pull up decky loader custom name shows up instead of the default name.   This would give the user just a bit more control over the UI.  Most of the plugins have self explanatory names but some are less so.  So the ability to change it could be useful.

robertefry:
This would be a really good addition.
+1

Leon-Swains:
I would also very much want this feature.

Waqam:
I would also very much like this feature, hacking it through the json file results in errors or broken plugins.

AAGaming00:
what is the usecase for this, i dont see it

PR #636, opened 2024-06-12
Added code for making it so the user can rename plugins.

oshata:
Please tick as appropriate:
- [ x] I have tested this code on a steam deck or on a PC
- [ ] This is a bugfix/hotfix
- [ x] This is a new feature

# Description
The intention is to make it so the user is able to rename plugins.  To do this an additional setting button was added in the settings \ plugins \ ...  context menu named "Rename".  This button should bring up a dialog with a text box, a enter button and a cancel button.  once the user has input the name they wish for the plugin they can push enter to rename it or cancel to exit out with out renaming it.

The files changed were:
backend\src\main.py
backend\src\plugin.py
frontend\src\components\settings\pages\plugin_list\index.tsx

This fixes issue: #629

Please provide a clear and concise description of what the new feature is. If appropriate, include screenshots or videos.
 I still have some testing to do the the code to fix some error.

### DIFF
--- a/backend/src/main.py
+++ b/backend/src/main.py
@@ -68,6 +68,7 @@ class PluginManager:
         self.plugin_browser = PluginBrowser(plugin_path, self.plugin_loader.plugins, self.plugin_loader, self.settings) 
         self.utilities = Utilities(self)
         self.updater = Updater(self)
+        self.web_app.add_routes([web.post('/plugins/{name}/rename', self.rename_plugin)])
 
         jinja_setup(self.web_app)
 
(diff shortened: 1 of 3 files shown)


TrainDoctor:
This PR fails all linter checks and appears to be LLM etc generated code. We cannot accept this PR as we do not accept LLM code and even if it wasn't, the complete failure to verify any of what the code actually does or verify it passes our linters is grounds enough for rejection. Please consider contributing through your own efforts. Thanks.

PR #661, opened 2024-07-18
Rewrite router/tabs/toaster hooks

AAGaming00:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This fixes many issues with the current router/toaster/tabs hooks (such as toasts breaking after exiting sleep), adds desktop BPM support to the router/toaster/tabs hooks, adds desktop support to the toaster hook, and adds preliminary support for a future desktop UI router hook.

The new toaster hook also supports displaying toasts in the QAM's notification tray.

Current blocking issues:
~~- [x] Investigate occasional react error when injecting into the router too early~~ I can't reproduce this anymore.
- [x] Make proper UI for toasts in the tray
- [x] Support desktop BPM in Navigation API (probably very easy)

Here's an example of toasts working in the tray (UI is WIP)
toasts in the QAM

# Potentially breaking changes
- Decky and plugins now load when the desktop Steam UI starts, as they would remain loaded after switching out of Big Picture anyway. I haven't observed any breakage due to this.
- Decky and plugin toasts now show in desktop mode


AAGaming00:
The tabs hook silently failed once when testing this, but i haven't been able to replicate it yet. Could use some help testing.

AAGaming00:
Should be ready for review and testing now.

backend/decky_loader/main.py
```diff
@@ -207,6 +207,7 @@ async def inject_javascript(self, tab: Tab, first: bool=False, request: Request|
                 await tab.close_websocket()
                 self.js_ctx_tab = None
                 await restart_webhelper()
+                await sleep(5)
```
suchmememanyskill, line 210: What does this sleep do? May be worth adding a comment so in the future we still know what this does

backend/decky_loader/utilities.py
```diff
@@ -195,7 +195,7 @@ async def http_request(self, req: Request) -> StreamResponse:
                 self.logger.debug(f"Finished stream for {url}")
         return res
 
-    async def http_request_legacy(self, method: str, url: str, extra_opts: Any = {}):
+    async def http_request_legacy(self, method: str, url: str, extra_opts: Any = {}, timeout: int | None = None):
```
suchmememanyskill, line 198: Unused argument?

suchmememanyskill (approved):
Python changes look fine to me.

backend/decky_loader/utilities.py
AAGaming00: intended to prevent kwargs error, no point actually implementing since its only for legacy compatibility but i might look into it

AAGaming00:
reported boot loops, this needs some work

backend/decky_loader/utilities.py
AAGaming00: implemented properly now

backend/decky_loader/main.py
AAGaming00: done

PR #676, opened 2024-08-09
[Rebase] Warn users on possible risks on installing third party unvetted plugins

RodoMa92:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This PR implements proper warning messages for third party install outside our store to warn users of potential malicious code outside our ecosystem.


frontend/src/components/modals/WarnThirdParty.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const [waitTimer, setWaitTimer] = useState(seconds);
+
+  useEffect(() => {
+    // exit early when we reach 0
+    if (waitTimer <= 0) return;
+
+    // save intervalId to clear the interval when the
+    // component re-renders
+    const intervalId = setInterval(() => {
+      setWaitTimer(waitTimer - 1);
```
AAGaming00, line 26: Wouldn't set timeout work better for this then? Unless I'm misunderstanding?

RodoMa92: Yeah, fair enough. Not really that skilled with react.

RodoMa92: Reading this now, no, because I'm referencing the timeout on the translation on line 58 to show the remaining time ticking down to give additional feedback to the user, so I can't see how I would be doing this with one single timeout tick compared to one for each second. It was like 6 months from when I wrote this. Let me know if I'm missing something.

AAGaming00:
can this be updated properly so we can finally merge it in

RodoMa92:
Yeah, sure, I'll take a look at it shortly :)

AAGaming00 (approved):
good will test soon

Issue #672, opened 2024-08-07
[Request] MOTD Display


beebls:
### Please confirm

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] This is not a request for a plugin

### Feature Request Description

In the event of plugin failures, necessary updates, or otherwise, Decky-Loader should have an Motd system to display messages in the frontend to users.

TrainDoctor:
A reference MOTD UI element exists on the branch: beebls/motd for the decky-loader side, but we now need our backend interface to interact with and send our Messages, OTD.

TrainDoctor:
> A reference MOTD UI element exists on the branch: beebls/motd for the decky-loader side, but we now need our backend interface to interact with and send our Messages, OTD.

This has now been added courtesy of @gbdlin for the server end and @AAGaming00 for our admin panel.
Now all that is needed is to PR this to main and have testers evaluate it.

PR #703, opened 2024-09-13
feat: Add announcements system

beebls:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This fixes issue: #672 

Adds an Announcements (formerly MOTD) display to the root QAM view,

The [plugins/testing].deckbrew.xyz/v1/-/announcements/current endpoint is used, and the latest announcement from this endpoint is display in the QAM.

Upon clicking the X button, that announcement's id is added to a hidden announcements list so that it isn't displayed again


frontend/src/components/AnnouncementsDisplay.tsx
```diff
@@ -0,0 +1,123 @@
...
+  Low: {
+    color: '#1488bb',
+    text: '#fff',
+  },
+};
+
+const welcomeAnnouncement: Announcement = {
+  id: 'welcomeAnnouncement',
+  title: 'Welcome to Decky!',
+  text: 'We hope you enjoy using Decky! If you have any questions or feedback, please let us know.',
```
suchmememanyskill, line 26: Where should people direct feedback?

(Also i assume this shows the first time you use decky? Might want to add something like start with looking at the store)

RodoMa92: That's a good point, but there is a blocking issue here. This strings need to be moves in en-US.json and not hardcoded here.

TrainDoctor (changes requested):
Remaining items before merge:
- [ ] Announcements must be checked for regularly similar to checks for updates to decky and for plugins.
- [ ] Markdown support for URLs etc.
- [ ] https://github.com/SteamDeckHomebrew/decky-plugin-store/issues/73
  - [ ] Store needs to expose modifiable expiration timeframe that is applied to all announcements by default.
- [ ] https://github.com/SteamDeckHomebrew/Plugin-Store-Admin/issues/1
  - [ ] Also needs to provide edit field for default announcement expiration timeframe.
- [ ] Announcement telling you which store you're on (optional but helpful)
  - [ ] Should be able to be disabled for frequent testers etc.
@beebls, @gbdlin and @AAGaming00: please make sure to coordinate with me so we can get this out asap.

AAGaming00:
how much of this still applies

PR #706, opened 2024-09-22
fix: Plugin install from ZIP getting stuck indefinitely

SilentException:
Please tick as appropriate:
- [X] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

This fixes issue: N/A

After debugging why plugin manual ZIP installations sometimes fail to finish and gets stuck I've noticed that plugin name is being extracted from folder name that is in the ZIP and not from the plugin.json file. This name is then used to search for plugin folder later on in the process (```plugin_folder = self.find_plugin_folder(name)```) and since the plugin folder name is pretty much guaranteed to be different than plugin name - assertion (```assert plugin_folder is not None```) is triggered.

This fixes it by reading the plugin name from plugin.json file that is included in the ZIP.
It might also be a good idea to check why the installer gets stuck when the assertion error is triggered but it is out of scope for this PR.

### DIFF
--- a/backend/decky_loader/browser.py
+++ b/backend/decky_loader/browser.py
@@ -213,7 +213,19 @@ class PluginBrowser:
                     return
 
                 else:
-                    name = sub(r"/.+$", "", plugin_json_list[0])
+                    plugin_json_file = plugin_json_list[0]
+                    name = sub(r"/.+$", "", plugin_json_file)
+                    try:
+                        with plugin_zip.open(plugin_json_file) as f:
+                            plugin_json_data = json.loads(f.read().decode('utf-8'))
+                            plugin_name_from_plugin_json = plugin_json_data.get('name')
+                            if plugin_name_from_plugin_json and plugin_name_from_plugin_json.strip():
+                                logger.info(f"Extracted plugin name from {plugin_json_file}: {plugin_name_from_plugin_json}")
+                                name = plugin_name_from_plugin_json
+                            else:
+                                logger.warning(f"Nonexistent or invalid 'name' key value in {plugin_json_file}. Falling back to extracting from path.")
+                    except Exception as e:
+                        logger.error(f"Failed to read or parse {plugin_json_file}: {str(e)}. Falling back to extracting from path.")
 
         try:
             pluginFolderPath = self.find_plugin_folder(name)


TrainDoctor:
@SteamDeckHomebrew/decky-testers priority on testing this one out so it can be released as part of v3.0.1!

eXhumer:
> This fixes it by reading the plugin name from plugin.json file that is included in the ZIP.

https://github.com/SteamDeckHomebrew/decky-loader/pull/578#pullrequestreview-1842550498

SilentException:
> > This fixes it by reading the plugin name from plugin.json file that is included in the ZIP.
> 
> [#578 (review)](https://github.com/SteamDeckHomebrew/decky-loader/pull/578#pullrequestreview-1842550498)
> 
> [#578 (comment)](https://github.com/SteamDeckHomebrew/decky-loader/pull/578#discussion_r1465647944)

Interesting, but it is not working as intended. The current code is deriving plugin ```name``` from folder name in the ZIP and then searching for that same ```name``` by traversing ```~/homebrew/plugins```and looking into ```plugin.json``` for the name.

SilentException:
Here are steps to reproduce and also test the fix.

You need GH actions run for new version of HLTB plugin: https://github.com/SteamDeckHomebrew/decky-plugin-database/actions/runs/10836675929
Not all plugins can trigger this bug (it depends on folder and plugin name).

It's a zip in a zip type scenario so unzip one layer first.
Try to manually install the ZIP. The installation will hang - not hang the system but hang as not finish, indefinitely - you can still cancel it. Plugin files will be extracted but noting will happen after that. After cancelling, you will not see plugin in the plugin list either (without restart)

Remove the previously installed plugin (if you want a clean state) and install the PluginLoader from this PR. Do the same manual ZIP installation - everything will be fine this time. Plugin will be installed, progress bar will move forward and Decky will show the new plugin in the list.

SilentException:
Anyone cares to take a look? I believe I gave good amount of info for this to be reviewed and tested?

backend/decky_loader/browser.py
```diff
@@ -213,7 +213,19 @@ async def _install(self, artifact: str, name: str, version: str, hash: str):
...
 
                 else:
-                    name = sub(r"/.+$", "", plugin_json_list[0])
+                    plugin_json_file = plugin_json_list[0]
+                    name = sub(r"/.+$", "", plugin_json_file)
+                    try:
+                        with plugin_zip.open(plugin_json_file) as f:
+                            plugin_json_data = json.loads(f.read().decode('utf-8'))
+                            plugin_name_from_plugin_json = plugin_json_data.get('name')
+                            if plugin_name_from_plugin_json and plugin_name_from_plugin_json.strip():
```
suchmememanyskill, line 222: Imo either use the stripped output or don't use strip at all. Do we need safeguarding against plugin.json's that have a space as their name?

suchmememanyskill (approved):
Tested on Steam version 1728093750 (beta), Decky 1781c19 + 819bd48 (Latest main + this PR). Installing CSS Loader via zip works as expected now

backend/decky_loader/browser.py
SilentException: Was meant to be a "null or witespace" or "blank" check.
First part of the condition will exclude missing `"name"` (None) and empty strings (`"name": ""`)
Second part will exclude blank strings (`"name": " "`).
Doesn't hurt I think, but for sure not needed. No problem to remove the stripped check.

suchmememanyskill: Was more so asking because ' my plugin ' would get through this check, but unsure if that's a good idea or would cause issues later. Either way also on the plugin dev imo if they enter something invalid in the name field.

SilentException: I understand. Plugin name `" my plugin "` wouldn't cause issues later (at least during this code path) because it is used in `find_plugin_folder` function that only does `if plugin["name"] == name:` and it will find the plugin folder without problems.

Maybe I should only check for None and not even try to deal with empty strings.
`if plugin_name_from_plugin_json is not None:`

PartyWumpus (approved):
(reviewing on behalf of testers i guess?) Works well, although at some point it would be nice if errors from here propagated up to the UI as toasts or something. Not a blocker, it was the same before too.
Steam version: 1726604483 (stable)

TrainDoctor:
Tested this myself and experienced no major issues during plugin install or during normal operation. Apologies for this taking so long to get merged.

TrainDoctor (approved):

PR #715, opened 2024-10-11
Fix the server-side tracking of updates

PartyWumpus:
Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [x] This is a new feature

# Description

Fixes the issue where every 'increment install count' request to the server was sent as a 'new download' even if it should've been 'update from existing install'. (the relevant bit is on line 198)

(Haven't actually tested it as i am doing this from my phone, will update when i get to my laptop)

### DIFF
--- a/backend/decky_loader/browser.py
+++ b/backend/decky_loader/browser.py
@@ -157,8 +157,16 @@ class PluginBrowser:
         # Will be set later in code
         res_zip = None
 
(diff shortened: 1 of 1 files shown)


TrainDoctor:
Tested locally and appears to work properly now. Merging so that this can go to pre-release and get more robust testing.

PR #733, opened 2024-12-24
feat: sync with local plugin status in store

alvaro-cuesta:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This fixes issue: #360

Adds some quality-of-life changes to the plugin store browser:

- Add ✓ icon to currently installed plugin version in version selector
- Change install button label depending on the install type that the button would trigger
- Adds icon to install button for clarity

The goal is to make it clear to the user what the current state of the installed plugin is, and what would be the impact of installing the selected version.

Screenshot

### DIFF
--- a/frontend/src/components/DeckyState.tsx
+++ b/frontend/src/components/DeckyState.tsx
@@ -138,7 +138,7 @@ export const useDeckyState = () => {
   }
 
   return deckyState;
-}
+};
 
 interface Props {
   deckyState: DeckyState;
--- a/frontend/src/components/DeckyState.tsx
+++ b/frontend/src/components/DeckyState.tsx
@@ -128,9 +128,17 @@ interface DeckyStateContext extends PublicDeckyState {
   closeActivePlugin(): void;
 }
 
-const DeckyStateContext = createContext<DeckyStateContext>(null as any);
+const DeckyStateContext = createContext<DeckyStateContext | null>(null);
 
-export const useDeckyState = () => useContext(DeckyStateContext);
+export const useDeckyState = () => {
+  const deckyState = useContext(DeckyStateContext);
+
+  if (deckyState === null) {
+    throw new Error('useDeckyState needs a parent DeckyStateContext');
+  }
+
+  return deckyState;
+}
 
 interface Props {
   deckyState: DeckyState;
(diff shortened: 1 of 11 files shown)


frontend/src/components/DeckyState.tsx
```diff
@@ -128,9 +128,17 @@ interface DeckyStateContext extends PublicDeckyState {
...
-export const useDeckyState = () => useContext(DeckyStateContext);
+export const useDeckyState = () => {
+  const deckyState = useContext(DeckyStateContext);
+
+  if (deckyState === null) {
+    throw new Error('useDeckyState needs a parent DeckyStateContext');
+  }
+
+  return deckyState;
+};
```
alvaro-cuesta, line 141: Note for reviewers: this tripped me up since the types assume this will never be `null`, which might not be true if `DeckyStateContext` is missing.

Opted to add this runtime check for some clarity on the error instead of failing later down the line.

frontend/src/components/PluginView.tsx
alvaro-cuesta, line 1: Note for reviewers: did this small refactor while exploring the code. Prevents some unnecessary re-renders.

I can post this as a separate PR if preferred.

frontend/src/components/store/PluginCard.tsx
```diff
@@ -1,18 +1,32 @@
...
+  storePlugin: StorePlugin;
+  installedPlugin: Plugin | undefined;
 }
 
-const PluginCard: FC<PluginCardProps> = ({ plugin }) => {
+const PluginCard: FC<PluginCardProps> = ({ storePlugin, installedPlugin }) => {
   const [selectedOption, setSelectedOption] = useState<number>(0);
-  const root = plugin.tags.some((tag) => tag === 'root');
+  const installedVersionIndex = storePlugin.versions.findIndex((version) => version.name === installedPlugin?.version);
+  const installType = // This assumes index in options is inverse to update order (i.e. newer updates are first)
```
alvaro-cuesta, line 18: Is this a correct assumption for the array we get from the store?

My first implementation had `OVERWRITE` for updates and downgrades too, but during testing I realized it was confusing when updates were available.

I considered using semver checks but I'm not sure if that's overkill if this assumption holds anyways, or whether the store forces semver at all, so I figured I'd just go with this simple assumption and discuss in PR.

frontend/src/components/DeckyState.tsx
AAGaming00: looks fine

TrainDoctor (approved):
Evaluated functionality, works great. Merging to main so users can give this a whirl on pre-release.

Issue #734, opened 2024-12-25
[Request] Allow opening hidden plugins


alvaro-cuesta:
### Please confirm

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] This is not a request for a plugin

### Feature Request Description

I have a bunch of plugins hidden in the quick access list because I seldomly access their panels. For example, plugin "ProtonDB Badges" is fire-configure-once-and-forget, so I don't want it to waste space and attention.

But this doesn't mean I will never want to access its configuration. Unless I'm missing something, currently the only way to access the plugin content panel is to unhide the plugin, access it in the quick access list, and then hide it again.

The requested feature is to have another way to access plugin panels without temporarily unhiding them.

### Further Description

I have considered multiple possibilities: 

1. Add a button/toggle to temporarily show all hidden plugins in the quick access list.

    I personally don't like the idea because this clutters the most common use-case (i.e. quickly accessing plugins).

2. Add an "Open" button inside Decky's settings, in each "Plugins" tab actions menu.

    When clicked the plugin content is shown in either:

    1. Replacing the plugin list content (as if navigating)

        I think this might have some issues like (1) the panels are meant to be shown in a narrow sidebar and (2) you can still open the sidebar and see a duplicate panel on top, which will have its own local state, so both panels will be shown at the same time but with desynced state.

    2. Inside a modal window.

        Has similar issues to option 1.

    3. Open the quick access sidebar, focused on the Decky tab, with the plugin already opened.

        I kinda like this solution (I have already coded a small prototype) even though it can get a bit weird due to how the sidebar can be opened/closed independently with the `...` physical button, focus contexts, handling of B back-navigation, etc. although I think it's tolerable and some can be mitigated (e.g. if we can listen to a "sidebar closed" event we could close the pre-opened plugin if deemed necessary).

    4. Some other option that I might have not considered?

    Additional considerations:

    1. I think it might make sense to replace the "Reinstall" button in the settings with the new "Open" button, and move "Reinstall" inside the actions menu. Feels like reinstalling a plugin is an even rarer action.

alvaro-cuesta:
I'm willing to work on this but first wanted to confirm that the feature makes sense to maintainers (I'd definitely use it and hide many more plugins than I currently have hidden!), that I didn't miss an existing way to do this, and what specific approach to implement.

[...]

Issue #350, opened 2023-01-22
[Request] Allow disabling plugins without uninstalling them


bspammer:
### Please confirm

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] This is not a request for a plugin

### Feature Request Description

Sometimes I want to stop using a plugin temporarily, but don't want to go through the hassle of uninstalling and then reinstalling it later, potentially resetting settings. It would be nice to have a toggle switch next to each plugin that allows disabling the plugin while leaving it installed.

bspammer:
For inspiration, this is how the toggle looks next to the plugins in the popular old school runescape client "Runelite":

![image](https://user-images.githubusercontent.com/11244353/213925005-94855f26-f0f4-448c-b71f-58a33fb7b1ab.png)

PartyWumpus:
plugin settings aren't deleted on uninstall and installation takes seconds so I personally don't think this is a worthwhile idea.

bspammer:
The actual installation itself is very quick, yes. Navigating to the store, and scrolling for a while to actually find the plugin is quite a lot of annoying overhead if it's something you do a lot. To give an example, I'll often uninstall powertools to compare the effect on fps that my current settings have with the steam deck defaults. Being able to do that with a single toggle would be really useful!

bspammer:
Also, if you haven't got any internet access then reinstalling isn't an option

PartyWumpus:
There's a new store UI coming out with a search bar soon, so it'll be faster to download. But yeah I suppose if you don't have internet it's not an option.

the example makes it clearer why you'd want that. I'm sure if you or someone else wants to make a PR to add this functionality the devs would be happy to take it but I think it's unlikely they'll dedicate time to it. (obviously I don't speak for them and you'll have to see what they think)

spudpiggy:
you shouldn't have to uninstall a plugin to disable it, even if you can just redownload it afterwards.
it just feels janky to not have the option, yk?
bumping for visibility

jcc10:
It would also be nice to have the option to "disable until next update" as in my situation, the deck UI got updated and now I am waiting on some plugins to update to the point where they don't completely break the UI.

I fully intend to use the plugins again later, I just don't want to have to keep hunting them down and checking if they got updated.

Managor:
I'm here for this feature. Sometimes plugins break due to changes and disabling a plugin would be nice to prevent error spam.

[...]

PR #736, opened 2024-12-28
feat: allow opening hidden plugins

alvaro-cuesta:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This fixes issue: #734 (see issue for alternatives considered)

This solves the use case where plugins are hidden from the quick access list because access to their panels is not used very often. For example, plugin "ProtonDB Badges" is fire-configure-once-and-forget, so it makes sense to hide it so it doesn't waste space and attention during quick access.

But this doesn't mean the user will never want to access hidden plugin panels to e.g. briefly change configuration. Currently the only way to access the plugin content panel is to unhide the plugin, access it in the quick access list, and then hide it again.

This allows opening plugins from the "Plugins" panel in Decky's settings, even if the plugin is hidden in the quick access list. Clicking the open button opens the plugin in the quick access sidebar, regardless of its hidden status.

Additionally:

- Since the "open" action is expected to be more common than the current "Reinstall" button (which seems to be there because it turns into "Update" when an update is available for the plugin):

    - Add "Open" action as the primary icon button
    - Move "Reinstall" option inside context menu so plugin can always be reinstalled (regardless of available updates)
    - Keep "Update" button in main interactables, but only show it when an update is available

    _(See [my comment below](https://github.com/SteamDeckHomebrew/decky-loader/pull/736#issuecomment-2564306625) for some thoughts around this.)_

- Add notification badge if plugin has updates

    Makes it more consistent with the rest of the app, which already has badges on available updates. It also makes it more obvious now that the "Update" button is only shown conditionally, calling attention to the updatable entries.

This is how the changes look in the "Plugins" settings tab:

Screenshot

### DIFF
--- a/backend/decky_loader/locales/en-US.json
+++ b/backend/decky_loader/locales/en-US.json
@@ -125,7 +125,6 @@
         "freeze": "Freeze updates",
         "hide": "Quick access: Hide",
         "no_plugin": "No plugins installed!",
-        "open": "Open",
         "plugin_actions": "Plugin Actions",
         "reinstall": "Reinstall",
         "reload": "Reload",
--- a/backend/decky_loader/locales/en-US.json
+++ b/backend/decky_loader/locales/en-US.json
@@ -125,6 +125,7 @@
         "freeze": "Freeze updates",
         "hide": "Quick access: Hide",
         "no_plugin": "No plugins installed!",
+        "open": "Open",
         "plugin_actions": "Plugin Actions",
         "reinstall": "Reinstall",
         "reload": "Reload",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -47,7 +47,7 @@
     }
   },
   "dependencies": {
-    "@decky/ui": "^4.8.3",
+    "@decky/ui": "^4.9.0",
     "compare-versions": "^6.1.1",
     "filesize": "^10.1.2",
     "i18next": "^23.11.5",
--- a/frontend/pnpm-lock.yaml
+++ b/frontend/pnpm-lock.yaml
@@ -9,8 +9,8 @@ importers:
   .:
     dependencies:
       '@decky/ui':
-        specifier: ^4.8.3
-        version: 4.8.3
+        specifier: ^4.9.0
+        version: 4.9.0
       compare-versions:
         specifier: ^6.1.1
         version: 6.1.1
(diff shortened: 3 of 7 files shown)


alvaro-cuesta:
Posted this PR because I'm running it on my fork and I figured having an actual runnable build would make sense to better discuss the ergonomics (see #734 for alternatives considered).

In particular, I'm not entirely sure if the "Open" action should be relegated to the context menu anyways, instead of having its own icon button.

I want to work on #350 soon-ish (which might deserve a toggle per row for conciseness, see [screenshot in first comment](https://github.com/SteamDeckHomebrew/decky-loader/issues/350#issuecomment-1399525121)) and want to leave a smooth path forward without too many UI changes, so it might make sense to not even add the "Open" as a main action in this PR to leave some space for the toggle.

- I expect 99% of the time users will just use the quick access menu, and for the very few cases that they really want to open a hidden plugin they can still reach through the context menu which is one click away.
- Seems like the toggle might be a useful action to have ready at hand, especially after updates that might break multiple plugins at once...
- ...but "open" seems useful as a main action too, so I can't make up my mind on which of the actions should stay (or both?)

I still think it makes sense to keep the "Reinstall" button tucked inside the context menu though (for the same space-saving reasons).

Some actual screenshots on how it could look like IF the plugins are disabled via toggle:

- This would be the "Plugins" tab with "Open" button AND disable toggle, which admittedly looks a bit busy and makes overflows more likely:

    With open button

- And this would be only the disable toggle, no "Open" button:

    No open button

- If the open action is tucked inside the context menu it could look like this (note that this uses my branch of SteamDeckHomebrew/decky-frontend-lib#115 which exposes `MenuSeparator`):

    Context menu

Thoughts?

TrainDoctor:
@alvaro-cuesta, hey sorry for the long time between when we last discussed this on discord and now.
@EMERALD0874 is our most versed UI/UX person on the team. I'll ask them to give this a review. Personally I love the toggle and open button in combination but I want Emerald's opinion if that's too much visually. I'm aware that a large portion of our users don't even know to open the settings menu until a plugin starts misbehaving so adding more interface might not be helpful.
Is it possible to make the open button only appear on plugins that have been hidden? If so I think that would be ideal imo.

alvaro-cuesta:
@TrainDoctor no worries, holidays are always a tough time, thanks for coming back to this.

> Is it possible to make the open button only appear on plugins that have been hidden?

The UX person at my job has this motto of better being consistent with what you show. E.g. in this case it might not be immediately obvious why some plugins have the buttons while others don't, which might puzzle some users.

But of course there are always exceptions and although I like dabbling in UX I'm not an expert either, let's see what @EMERALD0874 thinks!

alvaro-cuesta:
Now that SteamDeckHomebrew/decky-frontend-lib#115 is merged I added the menu separator to the plugin's menu (tested in both SteamOS Stable and Beta), splitting it in two sections:

- Install actions (reload, reinstall, uninstall)
- Status actions (hide, freeze)

Makes a bit less overwhelming and more structured, since this PR adds an additional menu item (and possibly more to come) with the "Reinstall" action, making the plugin list a bit more crowded. IMO it makes the context menu a bit more clear.

Normal mode:

Normal mode screenshot

Developer mode:

Developer mode screenshot

Even though the "normal" mode only has one status item I still think it's a good idea to guide the user through both sets of actions, plus SteamOS itself has multiple instances of guiding like this, e.g. see how "Switch to Desktop" is standing alone here to highlight it's its own kinda thing:

Power menu screenshot

EMERALD0874:
@alvaro-cuesta @TrainDoctor -  I'm sorry that I didn't see this until now. I wasn't requested as a reviewer and I think I have some weird settings for this repo with mentions, so it went under my radar.

I think both showing the plugin in the QAM and toggling them on/off should be put within the context menu, not a toggle. I imagine most users will want to have the plugins they installed consistently on, so making it possible to accidentally turn them off could be problematic compared to a button labeled "Enable" or "Disable" depending on the context. Maybe that could go beneath "Reload"?

AAGaming00:
stale, reopen if still interested

PR #739, opened 2025-01-02
 feat: add uninstall button to store 

alvaro-cuesta:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This fixes issue: NONE

This is a followup to PR #733. After discussing with @TrainDoctor on Discord he suggested it would be a good idea to also add an uninstall button to the store next to the "Reinstall" button (to plugins which are already installed).

Screenshot

To make the functionality more obvious, the button will only be shown along with "Reinstall", i.e. when the currently installed version is selected in the dropdown.

### DIFF
--- a/frontend/src/components/store/PluginCard.tsx
+++ b/frontend/src/components/store/PluginCard.tsx
@@ -1,7 +1,7 @@
-import { ButtonItem, Dropdown, Focusable, PanelSectionRow, SingleDropdownOption, SuspensefulImage } from '@decky/ui';
-import { CSSProperties, FC, useState } from 'react';
+import { DialogButton, Dropdown, Focusable, PanelSectionRow, SingleDropdownOption, SuspensefulImage } from '@decky/ui';
+import { FC, useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { FaArrowDown, FaArrowUp, FaCheck, FaDownload, FaRecycle } from 'react-icons/fa';
+import { FaArrowDown, FaArrowUp, FaCheck, FaDownload, FaRecycle, FaTrash } from 'react-icons/fa';
 
 import { InstallType, Plugin } from '../../plugin';
 import { StorePlugin, requestPluginInstall } from '../../store';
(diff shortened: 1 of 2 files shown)


frontend/src/components/store/PluginCard.tsx
```diff
@@ -32,7 +32,6 @@ const PluginCard: FC<PluginCardProps> = ({ storePlugin, installedPlugin }) => {
 
   return (
     <div
-      className="deckyStoreCard"
```
alvaro-cuesta, line 35: These classes seemed unused.

frontend/src/components/store/PluginCard.tsx
```diff
@@ -137,77 +128,82 @@ const PluginCard: FC<PluginCardProps> = ({ storePlugin, installedPlugin }) => {
...
-                className="deckyStoreCardInstallContainer"
-                style={
-                  {
-                    paddingTop: '0px',
-                    paddingBottom: '0px',
-                    flexGrow: 1,
-                    '--field-negative-horizontal-margin': 0,
-                  } as CSSProperties
-                }
+        <PanelSectionRow>
```
alvaro-cuesta, line 131: Sorry for the weird diff here (GitHub diff is not great) 🫤 This is just an unneeded `<div>` removed but the nesting removal is not shown very clearly here on GH. This is more clear in e.g. VSCode diff view.

frontend/src/components/store/PluginCard.tsx
```diff
@@ -137,77 +128,82 @@ const PluginCard: FC<PluginCardProps> = ({ storePlugin, installedPlugin }) => {
...
-                  selectedOption={selectedOption}
-                  onChange={({ data }) => setSelectedOption(data)}
-                />
-              </div>
-            </Focusable>
-          </PanelSectionRow>
-        </div>
+                <FaTrash />
+              </DialogButton>
+            ) : null}
```
alvaro-cuesta, line 186: Changed these `ButtonItem` into raw `DialogButton`. Is it okay?

As far as I can tell `ButtonItem` is geared towards standalone fields/actions on panels, but it has a lot of styling to support that use case which does not apply here. It was also not very style-able (was being a PITA to actually style properly here into a row of buttons with different sizes) and there was a lot of code around just trying to undo the field-like styling.

This seems to match the code style on the settings panel, but I'm not sure if I'm missing some reason to use `ButtonItem` in the store case.

TrainDoctor:
@AAGaming00 need you to confirm any questions PR submitter had to make sure there aren't any issues with frontend changes made. Once the frontend changes have been reviewed I'll give this a spin myself.

frontend/src/components/store/PluginCard.tsx
AAGaming00: this is probably fine it just used button item because store was implemented really early on

frontend/src/components/store/PluginCard.tsx
AAGaming00: they exist for theme developers, keep them

AAGaming00:
@alvaro-cuesta please revert 48f23f
the class names are there intentionally to aid theme devs

AAGaming00 (changes requested):
see comment

frontend/src/components/store/PluginCard.tsx
alvaro-cuesta: @AAGaming00 done. Also added a small warning documenting this.

PR #766, opened 2025-04-08
fix: swap Most/Least downloaded drop down in english store

xXJSONDeruloXx:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

Least and Most Downloaded sort options in English store appear to be swapped. I dont know for certain but I would think SteamGridDB has been downloaded way more than XIVOmega plugin.

This PR simply swaps the labels in the DD, but if these are mismatched for all other translations, we could consider adjusting the labels in store.tsx

image

### DIFF
--- a/backend/decky_loader/locales/en-US.json
+++ b/backend/decky_loader/locales/en-US.json
@@ -256,8 +256,8 @@
             "alph_desc": "Alphabetical (A to Z)",
             "date_asce": "Oldest First",
             "date_desc": "Newest First",
-            "downloads_asce": "Least Downloaded First",
-            "downloads_desc": "Most Downloaded First",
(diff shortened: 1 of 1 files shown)


EMERALD0874:
@xXJSONDeruloXx This is only a Band-Aid for a problem on the plugin store, not Decky Loader. Decky correctly asks for downloads in a descending order with the string "Most Downloaded First" presented to the user. The plugin store provides the incorrectly ordered response as seen [here](https://plugins.deckbrew.xyz/plugins?sort_direction=desc&sort_by=downloads).

Closing this as the fix should be on the [plugin store repo itself](https://github.com/SteamDeckHomebrew/decky-plugin-store), also fixing older Decky Loader versions.

Issue #768, opened 2025-04-11
[BUG] Reverse sort order in Decky Store


psxlover:
### Please confirm

- [x] I have searched existing issues
- [x] This issue is not a duplicate of an existing one
- [x] I have checked the [common issues section in the readme file](https://github.com/SteamDeckHomebrew/decky-loader#-common-issues)
- [ ] I have attached logs to this bug report (failure to include logs will mean your issue may not be responded to).

### Bug Report Description

When accessing the Decky Store, the order of the plugins seems to be the opposite of what is selected in "Sort" option. For example with

- `Alphabetical (A to Z)` the first plugin I see is vibrantDeck followed by steamdeck-input-disabler
-  `Alphabetical (Z to A)` the first plugin I see is "Animation Changer"
- `Least Downloaded First` starts with SteamGridDB
- `Most Downloaded First` starts with "XIVOmega" which is a plugin for a specific game, followed by "Decky Zerotier" which I've not seen before so I assume it's a relative new plugin and thus would have anong the lowest amount of downloads

### Expected Behaviour

Exactly the opposite. "Animation Changer" should be first in "A to Z" and I assume SteamGridDB is the most downloaded plugin, not the least downloaded one.

### SteamOS version

3.6.24

### Selected Update Channel

Stable

### Decky Loader Version

v3.1.5

### Plugin Info

- ProtonDB Badges - 1.1.0
- IsThereAnyDeal for Deck - 1.0.3
- Free Loader - 1.3.0
- HTLB for Deck - 2.0.4
- EmuDecky - 1.0.8
- SteamGridDB - 1.5.1-loaderv2
- Syncthing - 0.2.1
- Decky-Lookup - 0.1.0

### Backend Logs

Currently no backend logs

### Frontend Logs

Currently no frontend logs

RieLCho:
```
const dropdownSortOptions = useMemo(
    (): DropdownOption[] => [
      // ascending and descending order are the wrong way around for the alphabetical sort
      // this is because it was initially done incorrectly for i18n and 'fixing' it would
      // make all the translations incorrect
      { data: [SortOptions.name, SortDirections.ascending], label: t('Store.store_tabs.alph_desc') },
      { data: [SortOptions.name, SortDirections.descending], label: t('Store.store_tabs.alph_asce') },
      { data: [SortOptions.date, SortDirections.ascending], label: t('Store.store_tabs.date_asce') },
[...]

PR #781, opened 2025-06-13
Client side sorting/filtering for the Store page

LeonBlade:
Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This fixes issue: #768
**NOTE:** This fix doesn't change the backend and the original functionality of fetching with sort still has it's issue. 

Currently the way that the frontend sorts the plugin lists by going through the backend and making new network requests each time. This means wasted bandwidth and added latency for every time you want to change sorting order. Over time, more plugins will mean a bigger network response.

This PR introduces client side sorting and also filtering. The plugin list is fetched once when the component loads, and subsequent filtering and sorting is done on the client. This means we're only fetching data once on page load.

I also did expose a few missing properties for `StorePlugin` as part of this change and shouldn't affect functionality of the rest of the app.

I also did make a few adjustments to how the dropdown for sorting is handled with data opting for a singular key instead of the tuple. For filtering, I did opt for a pattern which allows for a more seamless integration of new filter patterns:
1. Add new filter to enum.
2. Create translation key matching the enum string value in the proper path.
3. Add functionality for the filtering in the `filterPlugin` function.

I'm interested in your thoughts on this implementation.

# Demo
image

### DIFF
--- a/backend/decky_loader/locales/en-US.json
+++ b/backend/decky_loader/locales/en-US.json
@@ -237,7 +237,12 @@
         },
         "store_filter": {
             "label": "Filter",
-            "label_def": "All"
+            "label_def": "All",
+            "options": {
+                "all": "All",
+                "installed": "Installed",
+                "not_installed": "Not Installed"
+            }
         },
         "store_search": {
             "label": "Search"
--- a/frontend/src/components/store/Store.tsx
+++ b/frontend/src/components/store/Store.tsx
@@ -19,7 +19,7 @@ import ExternalLink from '../ExternalLink';
 import PluginCard from './PluginCard';
 
 interface DropdownOptions<TData = unknown> extends SingleDropdownOption {
-    data: TData;
+  data: TData;
 }
 
 const logger = new Logger('Store');
--- a/frontend/src/components/store/Store.tsx
+++ b/frontend/src/components/store/Store.tsx
@@ -13,10 +13,14 @@ import { useTranslation } from 'react-i18next';
 
 import logo from '../../../assets/plugin_store.png';
 import Logger from '../../logger';
-import { Store, StorePlugin, getPluginList, getStore } from '../../store';
+import { SortKeys, Store, StoreFilter, StorePlugin, getPluginList, getStore } from '../../store';
 import { useDeckyState } from '../DeckyState';
 import ExternalLink from '../ExternalLink';
 import PluginCard from './PluginCard';
+
+interface DropdownOptions<TData = unknown> extends SingleDropdownOption {
+    data: TData;
+}
 
 const logger = new Logger('Store');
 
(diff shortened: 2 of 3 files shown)


sirwatt:
Any update on merging this? As of today sorting is still reversed, as mentioned on the original issue.

LeonBlade:
> Any update on merging this? As of today sorting is still reversed, as mentioned on the original issue.

They don't want to accept this as a fix to the issue due to the API being the core cause of this problem. I don't know if that code is public or not.

sirwatt:
> > Any update on merging this? As of today sorting is still reversed, as mentioned on the original issue.
> 
> They don't want to accept this as a fix to the issue due to the API being the core cause of this problem. I don't know if that code is public or not.

Went around searching and found this, but the problem is that it looks fine?
https://github.com/SteamDeckHomebrew/decky-plugin-store/blob/main/plugin_store/database/database.py#L224-L227
https://github.com/SteamDeckHomebrew/decky-plugin-store/blob/main/plugin_store/api/__init__.py#L145-L155

LeonBlade:
> > > Any update on merging this? As of today sorting is still reversed, as mentioned on the original issue.
> > 
> > 
> > They don't want to accept this as a fix to the issue due to the API being the core cause of this problem. I don't know if that code is public or not.
> 
> Went around searching and found this, but the problem is that it looks fine? https://github.com/SteamDeckHomebrew/decky-plugin-store/blob/main/plugin_store/database/database.py#L224-L227 https://github.com/SteamDeckHomebrew/decky-plugin-store/blob/main/plugin_store/api/__init__.py#L145-L155

Right now they flipped it on the front end. It's possible that things are just working fine now and they could accept this PR albeit with some changes. I'll have to look at it some more when I have some free time.

EMERALD0874:
I like many of the changes here, but sorting on the frontend could present a problem if pagination is ever added to the store. I believe that's why we have API requests when sorting.

LeonBlade:
> I like many of the changes here, but sorting on the frontend could present a problem if pagination is ever added to the store. I believe that's why we have API requests when sorting.

I believe someone else mentioned this before on the Discord. I'm happy to remove sorting logic from the frontend code.

sirwatt:
> I like many of the changes here, but sorting on the frontend could present a problem if pagination is ever added to the store. I believe that's why we have API requests when sorting.

Is this the right api url currently used? https://plugins.deckbrew.xyz/plugins?sort_by=name&sort_direction=asc  
Do you know why asc and desc are returning reversed?
Edit: oh and capitalized letters are showing up first, and lowercase at the end. in this case inverted because every response is inverted

EMERALD0874:
> Is this the right api url currently used? https://plugins.deckbrew.xyz/plugins?sort_by=name&sort_direction=asc Do you know why asc and desc are returning reversed? Edit: oh and capitalized letters are showing up first, and lowercase at the end. in this case inverted because every response is inverted

I've not had time to research the issue nor do I contribute to the plugin store repo, but the relevant code is [here](https://github.com/SteamDeckHomebrew/decky-plugin-store/blob/main/plugin_store/database/database.py#L204) if you'd like to take a look. I think the endpoint may've had issues sorting when I worked on the new store layout, leading to inconsistencies between the button label and API request made by Decky Loader.

LeonBlade:
At first glance I don't see anything wrong. I can't really get the plugin repo to work locally at the moment plus I have to seed it with some data. If someone knows more about issues with the API it would be better for now.

AAGaming00:
serverside is fixed now

PR #848, opened 2025-11-23
Add script to make debug info collection easier

jurassicplayer:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description
In an effort to make collection of debugging information easier for the end-user to obtain, I have written this script that will attempt to collect the OS name, OS version, steam client version, plugin information, backend logs, frontend logs, and archive them all into a single decky-debug-info.zip that users can just upload when adding or commenting on issues.

The script is VERY verbose about what it is doing, usually showing what commands it will run, and will attempt to scrub out the steam username from the frontend logs.

https://github.com/user-attachments/assets/4d4a04ff-522e-416f-aafd-717ead923b82


AAGaming00 (changes requested):
use env for bash shebang otherwise lgtm

AAGaming00 (approved):

PR #892, opened 2026-03-22
Fix setuid/setgid when running rootless

allmazz:
Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

Some plugins(even those that do not require root) fail when decky-loader has no setuid/setgid privileges. When the effective user is the same as host user, setuid and setgid calls must be skipped.

### DIFF
--- a/backend/decky_loader/localplatform/localplatformlinux.py
+++ b/backend/decky_loader/localplatform/localplatformlinux.py
@@ -116,28 +116,26 @@ def get_username() -> str:
     return _get_user()
 
 def setgid(user : UserType = UserType.HOST_USER):
-    user_id = 0
-
-    if user == UserType.HOST_USER:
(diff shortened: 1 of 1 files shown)


AAGaming00:
requires testing and rootless use is not supported so this may take us a while to merge but it looks fine to me

allmazz:
I'm running it rootless for about two weeks, and, surprisingly all plugins I tested worked fine. Even some of those that claim requirement of root permission to work.

AAGaming00:
the testing matters to make sure this doesn't break when running standard setups, testing on rootless does not matter

allmazz:
Of course, obviously it should be tested. What's the process? Maybe there is some kind of develop branch?

TrainDoctor (approved):

Issue #887, opened 2026-03-10
[Request] Add the ability to pin plugins to the QAM


LucaGPU:
### Please confirm

- [x] I have searched existing issues
- [x] This issue is not a duplicate of an existing one
- [x] This is not a request for a plugin

### Feature Request Description

The idea I have is the ability to pin any plugin i want to the QAM where the notifications, friends and decky tabs are 

### Further Description

This would be usefull for frequently used plugins if we could pin them to the QAM

hatim-cyber:
upvote

qd313:
This functionality would be great, especially for some people who want to quickly access a single plugin

PR #909, opened 2026-05-24
feat: allow pinning plugins to top-level QAM

devbdavis:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This fixes issue: #887.

Adds an option to pin any installed plugin from the Decky tab up to the top level of the Quick Access Menu, so it gets its own QAM tab above the Decky plug icon instead of being nested under it. Use case from the issue: jumping straight to a music controller plugin without going through the Decky tab.

- New per-plugin **Pin directly to Quick Access Menu** / **Unpin from Quick Access Menu** action in the plugin list context menu (Decky → gear → Plugins → ⋯).
- Pinned plugins get their own QAM tab placed above the Decky tab; the tab renders the plugin's `titleView` (or its name) as a heading, then its content — matching the inner Decky tab plugin view.
- Pinned plugins are hidden from the Decky tab's plugin list so the UI isn't duplicated.
- Tab ordering respects the existing `pluginOrder` from the plugin list (drag-reorder applies on next QAM open — same lifecycle as every other tab change).
- State persisted under a new `pinnedPlugins` setting; cleared on uninstall in `cleanup_plugin_settings`, and the frontend service is invalidated from the uninstall modal alongside the existing frozen/hidden invalidations.
- `tabs-hook` extended to support tab removal and stable ordering relative to the Decky tab (previously it was add-only).

## Implementation notes

- New `PinnedPluginsService` mirrors the existing `HiddenPluginsService` / `FrozenPluginService` pattern (settings get/set + invalidate).
- `DeckyState` gains a `pinnedPlugins: string[]` field with setter + publicState entry.
- `PluginLoader` subscribes to deckyState updates and reconciles QAM tabs via `tabsHook`. Each pinned plugin gets a stable numeric tab id in `[100, 998]` derived from its name (with collision rehash) so it sits between Steam's built-in tabs and `QuickAccessTab.Decky` (999).
- `PinnedPluginView` resolves the plugin via `useDeckyState`, so plugin reloads/dev hot-reloads don't strand the tab; while the plugin is unloaded the tab is just empty rather than crashing.
- Only `en-US.json` strings are added — other locales fall back until Weblate contributors translate, matching how prior strings like `freeze`/`hide` were introduced.

## Test plan

Verified end-to-end on a ROG Ally X running Bazzite + dev decky-loader. Existing decky behaviors (hide, freeze, reorder, uninstall) still work; build/typecheck/prettier all clean.

- [x] Pin a plugin from Decky → gear → Plugins; close/reopen QAM → plugin appears as its own tab above the Decky plug icon with its name as a heading
- [x] Pinned plugin no longer appears in the Decky tab's plugin list
- [x] Unpin the same plugin → standalone tab disappears, plugin returns to the Decky tab list
- [x] Drag-reorder plugins in settings → tab order in QAM reflects the new order on next open
- [x] Uninstall a pinned plugin → no orphan tab on next launch; `pinnedPlugins` setting cleared
- [x] Pin multiple plugins → all stack above Decky in the configured order
- [x] Plugin label in settings shows the "Pinned" indicator while pinned

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/decky_loader/browser.py
+++ b/backend/decky_loader/browser.py
@@ -341,6 +341,11 @@ class PluginBrowser:
             hidden_plugins.remove(name)
             self.settings.setSetting("hiddenPlugins", hidden_plugins)
 
+        pinned_plugins = self.settings.getSetting("pinnedPlugins", [])
+        if name in pinned_plugins:
+            pinned_plugins.remove(name)
+            self.settings.setSetting("pinnedPlugins", pinned_plugins)
+
         plugin_order = self.settings.getSetting("pluginOrder", [])
 
         if name in plugin_order:
--- a/backend/decky_loader/locales/en-US.json
+++ b/backend/decky_loader/locales/en-US.json
@@ -126,12 +126,14 @@
         "freeze": "Freeze updates",
         "hide": "Quick access: Hide",
         "no_plugin": "No plugins installed!",
+        "pin": "Pin directly to Quick Access Menu",
         "plugin_actions": "Plugin Actions",
         "reinstall": "Reinstall",
         "reload": "Reload",
         "show": "Quick access: Show",
         "unfreeze": "Allow updates",
         "uninstall": "Uninstall",
+        "unpin": "Unpin from Quick Access Menu",
         "update_all_one": "Update 1 plugin",
         "update_all_other": "Update {{count}} plugins",
         "update_to": "Update to {{name}}",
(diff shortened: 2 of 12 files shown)


AAGaming00:
ai

devbdavis:
Please consider a no-ai pledge on your readme to deter users like me. This change was thoroughly reviewed by a human and tested on device before offering it. I use this for the Music Control plugin to make it top level and it works great.

But I totally understand the stance. Apologies again and wishing you the best. Thank you for making and maintaining such a cool tool!

PR #913, opened 2026-06-03
feat: add backend Mac support

giodotblue:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [X] This is a new feature

# Description

`localplatformlinux.py` mostly works on Mac but fails when it hits systemd stuff

This adds `localplatformmac.py` which stubs a few things and uses the Linux implementations for everything else

image

### DIFF
--- a/backend/decky_loader/localplatform/localplatform.py
+++ b/backend/decky_loader/localplatform/localplatform.py
@@ -1,11 +1,15 @@
 import platform, os
 
 ON_WINDOWS = platform.system() == "Windows"
-ON_LINUX = not ON_WINDOWS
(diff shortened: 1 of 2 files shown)


AAGaming00 (approved):
but why tho

giodotblue:
Haha I'm using it for this https://x.com/giodotblue/status/2062893278891278504?s=20

It's a pretty small change, I think it's better to have a separate file instead of just gating off the systemd calls in localplatformlinux though

AAGaming00:
damn i was trying to do that ages ago

PR #927, opened 2026-06-24
feat: add OpenRC (Gentoo) + svcmgr check (Linux)

NyaomiDEV:
Please tick as appropriate:
- [ ] This is a bugfix/hotfix
- [x] This is a new feature

# Description

This PR adds two new services for OpenRC usage - at least on my HTPC system (<mod snip - Arch-based distro-that-shall-not-be-named>).

Gentoo people may want to test, as well as proper Systemd users, as I am de-Systemd-ed and I don't know if I broke it for y'all (though I shouldn't really have.)

I have to say, this is NOT AI coded but I suck at Python so if I missed something do tell me, as that would be a genuine "doesn't know how to Python" error.

### DIFF
--- a/backend/decky_loader/localplatform/localplatformlinux.py
+++ b/backend/decky_loader/localplatform/localplatformlinux.py
@@ -10,7 +10,7 @@ logger = logging.getLogger("localplatform")
 
 def _get_service_manager() -> str:
     try:
-        if os.getenv("LINUX_SERVICE_MANAGER"): return os.getenv("LINUX_SERVICE_MANAGER")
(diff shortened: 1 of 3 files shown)


AAGaming00 (approved):
from my cursory understanding of openrc like 4 years ago these look fine to me

NyaomiDEV:
I wanted to add that one liner to override service manager definition - could be useful for debugging or for people who have niche configurations

NyaomiDEV:
If this doesn't regress Systemd, I think it's good to merge it

NyaomiDEV:
fwiw it's been some days with this in my htpc and it works without a hitch

AAGaming00:
works fine

AAGaming00 (approved):

PR #935, opened 2026-07-03
fix: ignore whitespace in store plugin search

### LINKED ISSUE #622: [Request] Make the plugin store fuzzy sort, or at least strip out whitespace
### Please confirm

- [X] This is not a request for a plugin

### Feature Request Description

Title. Plugin searching is sometimes a little flaky, for example if you search for `tab master` you won't find it because it's `tabmaster`. This case can be avoided by ignoring all whitespace, but it might be better to get a proper fuzzy sort on the store.

This issue should technically be on the decky-store repo I think, as it'll likely be implemented there.

### Further Description

_No response_

zdennisz:
Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

The plugin store search used plain substring matching, so a query like `tab master` would not match a plugin named `tabmaster`. This normalizes both the search query and the searched fields (name, description, author, tags) by lowercasing and stripping whitespace before comparing.

Store search is performed entirely client-side (the remote store API returns the full list; only `sort_by`/`sort_direction` are sent), so this fix is contained to `frontend/src/components/store/Store.tsx`.

### Verification
- `pnpm run typecheck` — clean
- prettier — clean
- Logic check: `tab master`, `tabmaster`, `css`, `cssloader` all match `TabMaster`/`CSS Loader`; empty query still returns all plugins.

Resolves #622

### DIFF
--- a/frontend/src/components/store/Store.tsx
+++ b/frontend/src/components/store/Store.tsx
@@ -20,6 +20,10 @@ import PluginCard from './PluginCard';
 
 const logger = new Logger('Store');
 
+// Normalize a string for searching: lowercase and strip all whitespace so that
+// queries like "tab master" match plugins named "tabmaster" (see issue #622).
+const normalizeForSearch = (value: string): string => value.toLowerCase().replace(/\s+/g, '');
+
 const StorePage: FC<{}> = () => {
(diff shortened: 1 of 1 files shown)


AAGaming00:
no ai in loader

PR #942, opened 2026-07-14
Upstream action workflows

RodoMa92:
Updated older action workflows and adapted the with variables where needed.

Python has been also bumped to the latest 3.11 LTS release, but we will probably need to consider a mayor version bump in the future, since it has only ~ 1 year of security updates.


ynhhoJ:
Good work! Really necessary one!

Issue #943, opened 2026-07-15
[BUG] Progress bar layout breaks in the confirmation dialog during plugin installs/updates


beallio:
### Please confirm

- [x] I have searched existing issues
- [x] This issue is not a duplicate of an existing one
- [x] I have checked the [common issues section in the readme file](https://github.com/SteamDeckHomebrew/decky-loader#-common-issues)
- [x] I have attached logs to this bug report (failure to include logs will mean your issue may not be responded to).

### Bug Report Description

When installing, reinstalling, or updating a plugin, the progress info in the confirmation dialog gets laid out pretty weirdly. It looks like the progress widget is relying on an inline, fixed-width layout. Because of this, the operation text gets shoved to the right, the progress bar shifts (or just straight-up overflows) whenever the text changes, and the whole panel fails to line up with the dialog buttons below it.

---

### **Steps to Reproduce**

1. Open the Decky plugin store.
2. Trigger an install, reinstall, or update for any plugin.
3. Confirm the operation.
4. Watch the progress text and bar while it runs.

<img width="640" height="360" alt="Image" src="https://github.com/user-attachments/assets/c9cd3ebd-3659-45b3-a33f-9427d7476392" />

### Expected Behaviour

* The progress panel should sit neatly below the description and span the full width of the modal content (aligning with the buttons below).
* The operation text should be left-aligned.
* The progress bar needs to fill the panel while keeping its standard internal padding.

### SteamOS version

SteamOS 3.8.14

### Selected Update Channel

Stable

### Decky Loader Version

v3.2.6

### Plugin Info

```
Directory: /home/deck/homebrew/plugins/CheatDeck/
Package Name: cheatdeck
Version: 1.2.1-ffa3314
-----------------------------
Directory: /home/deck/homebrew/plugins/Decky-Metadata/
Package Name: decky-metadata
Version: 0.2.1+8b7b1ac
-----------------------------
Directory: /home/deck/homebrew/plugins/decky-steamgriddb/
Package Name: decky-steamgriddb
Version: 1.7.1
-----------------------------
Directory: /home/deck/homebrew/plugins/decky-storage-cleaner/
Package Name: decky-storage-cleaner
Version: 1.4.0
-----------------------------
Directory: /home/deck/homebrew/plugins/decky-syncthing/
Package Name: decky-syncthing
Version: 0.2.7
-----------------------------
Directory: /home/deck/homebrew/plugins/hltb-for-deck/
[...]

PR #944, opened 2026-07-15
fix: align plugin install progress

beallio:
Please tick as appropriate:

  - [x] I have tested this code on a Steam Deck or on a PC
  - [x] This is a bugfix/hotfix
  - [ ] This is a new feature

  # Description

  Moves plugin installation progress out of the dialog title and into a shared, full-width progress panel below the description.

  Previously, the progress widget used an inline, fixed-width layout that pushed the operation text to the right and caused the progress bar to overflow or shift as the text changed.

  The new progress panel:

  - Spans the modal content width, aligning with the outer edges of the dialog buttons.
  - Left-aligns the operation text.
  - Keeps the progress bar within the panel’s standard padding.
  - Uses Decky’s native Field and ProgressBar components.
  - Shares the same layout between single-plugin and multi-plugin installations.

Screencast_20260715_130437

  This fixes issue: #943


AAGaming00 (approved):
lgtm

PR #952, opened 2026-08-13
Fixed blank listing with file picker

DoomTay:
Typically happened when starting with a file

Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

This fixes issue an issue where if the starting path was a file, the file listing would be blank until you went up a level


frontend/src/components/modals/filepicker/index.tsx
```diff
@@ -193,6 +193,7 @@ const FilePicker: FunctionComponent<FilePickerProps> = ({
           page,
           max,
         );
+        if (path != listing.realpath) setPath(listing.realpath);
```
bricefriha: we may as well create a constant for the path since we are using it twice

```suggestion
       const listingPath = listing.realpath;
        if (path != listingPath) setPath(listingPath);
```

DoomTay: I never thought of consts, let alone variables being handled that way, especially an attribute of what is already a const

bricefriha: you never thought of using a const use a value twice?

DoomTay: I meant assigning an attribute of a const to another const

frontend/src/components/modals/filepicker/index.tsx
```diff
@@ -193,6 +193,8 @@ const FilePicker: FunctionComponent<FilePickerProps> = ({
           page,
           max,
         );
+       const listingPath = listing.realpath;
```
ynhhoJ: Use proper indentation here

PR #962, opened 2026-09-10
fix: restore fullscreen pages on newer Steam clients

virtudude:
Please tick as appropriate:
- [x] This is a bugfix/hotfix
- [ ] This is a new feature

# Description

Allow multi-character identifiers in the Route and error-reporting getter matchers so Decky can discover both hooks on newer clients. Preserve the existing legacy matching forms.

Fix was verified on ARM64 Steam client version `1788989629` running on Armada.


AAGaming00 (approved):
looks sane will test shortly

zkrvf:
Confirmed working on an Steam Machine running SteamOS 3.8.26 with Decky v3.2.8.
After applying this PR, Store, Settings, and plugin pages load correctly again. The previous Route and ErrorBoundary errors also disappeared from cef_log.txt. 

Thanks!

roblesi:
Confirming #962 fixes this on an x86_64 non-Deck client, as a second data point alongside @zkrvf's ARM64 Steam Machine report.

**Environment**
- CachyOS (Arch), kernel `7.1.8-1-cachyos-deckify`, x86_64 desktop, Gamepad UI under gamescope
- Steam `steamdeck_publicbeta`, package CL `1789086785`
- Decky `v3.2.8-pre1` (the same matchers are present in `v3.2.8`)

**Root cause**

The Steam UI chunk that carries the router (`steamui/chunk~2dcc5aaf7.js`, mtime 2026-09-10 18:22) now emits multi-character minified identifiers, so the single-`.` matcher cannot match:

```console
$ grep -cE 'routePath:.\.match\?\.path.' chunk~2dcc5aaf7.js
0
$ grep -oE 'routePath:[^,]{0,24}' chunk~2dcc5aaf7.js
routePath:ve.match?.path
routePath:q
```

`ve`, not a single character. `RouterHook.Route` therefore stays `undefined`, `processList()` never builds any route elements, and every Decky route renders an empty shell: `/decky/settings` and `/decky/store` were both 16,582-byte bodies against 132,952 for `/library/home`.

Worth flagging that this fails *silently*... no exception is thrown anywhere, which is why the result is a blank page rather than an error boundary, and why reinstalling looks like it does nothing.

The `errorboundary-hook.tsx` matcher breaks the same way on this client, so that hook disabled itself early. A side effect is that `_deckyForceRerender` is absent from the tree, which also makes the forced re-render at the end of `patchGamepadRouter()` a no-op.

**After applying #962** (both regex changes, nothing else), live state in `SharedJSContext`:

| | before | after |
|---|---|---|
| `routerHook.Route` | `undefined` | `function` |
| `routerHook.routes` | `null` | 32 |
| `errorCheckPatch` / `errorBoundaryPatch` | absent | both installed |
| `_deckyForceRerender` present in tree | no | yes |

Settings, Store (all 110 plugins listed) and plugin pages all render again.

One note for anyone arriving from search: **switching Steam back to Stable does not help**, the bundle change is present on all branches, as @Hatchetforce and @Jack32X both found above.

Hatchetforce:
@roblesi I was just on my Steam Machine. I opened Decky, which had been working normally for Stable, Beta, and Preview. There was an update available. I applied it. It immediately broke Decky functionality for the Beta Channel. Returning to the Stable Steam OS Channel restores all Decky functionality. 

EDIT: The update reinstalls the current release of Decky 3.2.8 and breaks the Beta branch of Steam OS. Reinstallation of the test fix #962 resolves Decky functionality for the Beta Branch.

roblesi:
@Hatchetforce ah, good catch! thanks for coming back with this!

That "does not help" line of mine came from your earlier comment over in #964 ("Switching to Stable, Beta, Preview branches of Steam OS … presents the same symptoms") plus @Jack32X's, rather than from a rollback I'd actually tested. Everything else in my comment came off my own machine, but that bit didn't, so it's only fair for me to walk it back.

Here's a thought on why we might both be right: the SteamOS channel (Stable/Beta/Preview) and the Steam *client* branch (`package/beta`, e.g. `steamdeck_publicbeta`) are two separate knobs, and a non-SteamOS install like mine only has the second one. What really decides things is which `steamui/` bundle you end up running. Land on a combination still serving a build from before the multi-char minification and Decky is perfectly happy; once that branch catches up to the newer bundle, the blank pages come right back. That would fit it working for you today but not on the 10th, and other folks seeing no difference at all.

So I'd think of Stable as a reprieve rather than a cure, it holds until your branch catches up, which is more or less what your Decky update just demonstrated. #962 looks like the genuine fix, and it's held up nicely on my client build.

One thing worth flagging for anyone else landing here: Decky's updater reinstalls stock 3.2.8 and quietly undoes the patch, which is exactly what bit you. Worth re-applying #962 after any Decky update until this merges.

Hatchetforce:
@roblesi  First let me say I am sorry if I implied any disparagement for your work or comments. I feel like a mental midget here :) next to your comments. You have far more understanding of this than I. All I can do is report symptoms, but you possess an innate comprehension of cause and effect. 

That said, I think you are right. As I posted earlier, despite Stable working normally, that is not sufficient and it is evident reapplication of #962 resolves a lot issues including Beta and Preview Steam OS Branches. So the key in this case lies in what effective matters vary between Stable Steam OS and Beta Steam OS. That is going to take us to the exact root and cause. 

Cheers my friend, Your deductive work is amazing.

Ciao.

roblesi:
@Hatchetforce omg no apology needed! All good!

beallio:
Are these comments all from Claude bots?

roblesi:
> Are these comments all from Claude bots?

uh, no xDD

Hatchetforce:
> Are these comments all from Claude bots?

Bwahahaha. Claude was too busy after we screwed over its plans for August 29th.

ordinarylinuxuser:
@Hatchetforce #962 will not fix all the issues, proton db badges and some other plugins are broken still. as discussion on the discord i think there is still some more fixes needs to be done.

moi952:
Hi, am I the only one unable to install a ZIP plugin?
This fix resolved the main issue, but it doesn't allow installing or updating a ZIP plugin.

Thanks

Hatchetforce:
> Hi, am I the only one unable to install a ZIP plugin? This fix resolved the main issue, but it doesn't allow installing or updating a ZIP plugin.
> 
> Thanks

I don't mean to ask a stupid question but did you enable developer mode?

moi952:
> > Hi, am I the only one unable to install a ZIP plugin? This fix resolved the main issue, but it doesn't allow installing or updating a ZIP plugin.
> 
> > 
> 
> > Thanks
> 
> 
> 
> I don't mean to ask a stupid question but did you enable developer mode?

It was already enabled before. 
However, to test it, I downloaded the artifact from the build (PluginLoader) and placed it in `homebrew/services`, but I have to switch back to Desktop Mode for Decky Loader to launch. 

Is that normal? 

Or did I install this test version incorrectly?

Hatchetforce:
After installation, you should not have to switch to desktop mode for Decky to launch. It has a history of not working right under Desktop anyway unless you're in Big Picture Mode under Desktop which is different than gaming mode, and you should be able to launch it normally in gaming mode. 

As soon as you are up in gaming mode you should be able to hit the quick access button and Decky should be there. Then go into the settings for Decky and enable the developer mode so you can install plugins like Mako from zip files.

I tried to get it to work correctly with a hand installation as well and still had issues. The only way I could get it to function somewhat normally was to go to the stable branch of Steam OS, and a fresh install of Decky. Then apply the 962 fix from inside Decky. Then I could go back to the beta branch of steam and it would still work, somewhat. There are still issues with certain plugins not functioning correctly under Decky. This is all tied to the latest stable, beta branch, and preview branch of steamos. They've made some changes that are generating these issues.

moi952:
> After installation, you should not have to switch to desktop mode for Decky to launch. It has a history of not working right under Desktop anyway unless you're in Big Picture Mode under Desktop which is different than gaming mode, and you should be able to launch it normally in gaming mode. 
> 
> 
> 
> As soon as you are up in gaming mode you should be able to hit the quick access button and Decky should be there. Then go into the settings for Decky and enable the developer mode so you can install plugins like Mako from zip files.
> 
> 
> 
> I tried to get it to work correctly with a hand installation as well and still had issues. The only way I could get it to function somewhat normally was to go to the stable branch of Steam OS, and a fresh install of Decky. Then apply the 942 fix from inside Decky. Then I could go back to the beta branch of steam and it would still work, somewhat. There are still issues with certain plugins not functioning correctly under Decky. This is all tied to the latest stable, beta branch, and preview branch of steamos. They've made some changes that are generating these issues.

I don't understand, then. Decky doesn't launch on startup, I have to switch back to the desktop. I know how to install plugins—I develop them myself. 

However, I was talking about installing the fix, do I need to switch back to the Stable Steam channel?

Hatchetforce:
Ah okay. If decky doesn't launch or doesn't show up properly switching back to the stable channel should allow it to launch. 

At that point go into decky and install the 962 fix from the menu. Then you can go back to the beta steam channel. 

But be aware you will get a notification that Decky has an update. Don't install it. If you do it's going to simply overwrite the fix with the latest build and you'll be back to the problem again. If that happens then just switch back to the stable channel launch decky reinstall the fix and then back to the beta channel

Hatchetforce:
> @Hatchetforce #962 will not fix all the issues, proton db badges and some other plugins are broken still. as discussion on the discord i think there is still some more fixes needs to be done.

Correct. 962 does not fix all the issues. Quite a few remain. SteamGrid is another that is still throwing errors.

afk-rightNow:
The remainder of the issue seems to come from missing resolution of the `ProgressBar` UI component. Looks like Steam might have altered the export leaving the current dynamic matcher throwing an undefined. 

This might help
```typescript
export const ProgressBar = findModuleExport((entry: Export) => {
  const source = entry?.toString?.();

  return (
    source?.includes('.ProgressBar,"standard"==') ||
    (
      source?.includes('role:"progressbar"') &&
      source?.includes('"aria-valuenow"') &&
      source?.includes('nProgress') &&
      source?.includes('.ProgressBar') &&
      source?.includes('.Indeterminate')
    )
  );
}) as FC<ProgressBarProps>;

Hatchetforce:
Just tested updated decky on steam machine with the beta branch which also just updated to 3.8.27.

Fix is functional as before when applied separately for quite a few plugins however there are still issues with a couple including steam grid.

